PR #16, opened 2015-04-07
Once again enable build of qmlwidgets

arnost00:
Because all of QML modules import this module.
So they didn't load when qmlwidgets are disabled from build.
Tested only on Windows.

### DIFF
--- a/libqf/plugins/plugins.pro
+++ b/libqf/plugins/plugins.pro
@@ -3,7 +3,7 @@ CONFIG += ordered
 
 SUBDIRS += \
 	core \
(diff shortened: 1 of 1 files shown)


fvacek:
qmlwidgets plugin is not used any more, so if there is still reference to this plugin in quickevent code, it should be cleared there. There is no other reason to enable not used QML plugin.

fvacek:
There is dependency in PluginManifest, so qmlwidgets plugin should be enabled to satisfy this dependency, all other references are obsolete for now.

Issue #54, opened 2015-09-01
Time loss information on receipt ticket is sometimes incorrect


fvacek:
It seems that it will happen when SI station is broken or when some competitor miss some punches. More investigation is needed.

fvacek:
FIXED

PR #56, opened 2015-09-14
Fix of bad loss times on receipts

mezkov:
This fixes printing bad loss time on receipts after missing punch. (issue #54 )


fvacek:
Great thanks :)

PR #110, opened 2016-04-02
Edited translations

Paukert:


fvacek:
Lukasi super, diky moc.

PR #154, opened 2016-05-25
eventplugin: Fix SQL statement

lpechacek:
Commit 78865b742dd71 broke event setup sequence which manifested itself with
following messages in application debugging output:
  88<WARN>[query.cpp:30] "2 values for 1 columns Unable to execute statement"
  89<WARN>[query.cpp:60] " Parameter count mismatch"

It is disturbing that this error passed unnoticed.


fvacek:
Thanks :)

PR #176, opened 2016-06-28
Coverity scan fixes

lpechacek:
Coverity has identified several instances of coding errors in QuickEvent code.  The most serious were uninitialized class members.  This patch set contains both plain fixes and code cleanup.  Please review.

https://scan.coverity.com/projects/quickbox

### DIFF
--- a/libqf/libqfqmlwidgets/src/framework/mainwindow.h
+++ b/libqf/libqfqmlwidgets/src/framework/mainwindow.h
@@ -99,7 +99,6 @@ private:
 	//void setupSettingsPersistence();
 private:
 	PluginLoader *m_pluginLoader;
-	qf::qmlwidgets::StatusBar* m_statusBar;
 	QMap<QString, qf::qmlwidgets::ToolBar*> m_toolBars;
 	static MainWindow *self;
 };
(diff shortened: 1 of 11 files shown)


libqf/libqfqmlwidgets/src/tableview.cpp
```diff
@@ -2067,12 +2067,12 @@ void TableView::removeSelectedRowsInline()
 
 bool TableView::edit(const QModelIndex& index, EditTrigger trigger, QEvent* event)
 {
-	qfLogFuncFrame() << "trigger:" << trigger << "event type:" << ((event)? event->type(): 0) << "editTriggers:" << editTriggers();
```
fvacek: how do you know that event is always not NULL

fvacek:
Good job, thanks a lot :)

libqf/libqfqmlwidgets/src/tableview.cpp
lpechacek: Good point.  Although Qt documentation doesn't say anything about event pointer value, grep through Qt sources shows that CurrentChanged trigger comes with NULL event pointer.  That means these checks must stay in place and event->accept() can be called without any checking as it's guarded by the trigger type test.  I'll drop this patch from the series.

lpechacek:
Thanks for the review!  The defective patch has been dropped.

PR #232, opened 2016-12-16
Coverity scan fixes

lpechacek:


libqf/libqfcore/src/sql/connection.cpp
```diff
@@ -174,7 +174,7 @@ static QVariant sqlite_set_pragma(QSqlQuery &q, const QString &pragma_key, const
 		qfError() << QString("SQL Error\nquery: %1;").arg(qs);
 		return QVariant();
 	}
-	q.next();
+	QF_CHECK(q.next(), "SQL PRAGMA query is supposed to return data but it hasn't");
```
fvacek: please use rather
```c++
if(q.next()) {
...
}
else {
  qfError() << "SQL PRAGMA query is supposed to return data but it hasn't";
}
```

quickevent/app/plugins/qml/Event/src/Event/eventconfig.cpp
```diff
@@ -106,13 +106,13 @@ void EventConfig::save(const QString &path_to_save)
 			val_str = val.toString();
 		q_up.bindValue(":key", key);
 		q_up.bindValue(":val", val);
-		q_up.exec();
+		q_up.exec(qf::core::Exception::Throw);
```
fvacek: `q_up.exec(!qf::core::Exception::Throw);` 

is intention here, there is no try-catch block in function call stack, so I prefer not crashing app with error in log in this case

If we want to enable throw exception here, we should change more code in other places

libqf/libqfcore/src/sql/connection.cpp
lpechacek: OK, done.

quickevent/app/plugins/qml/Event/src/Event/eventconfig.cpp
lpechacek: Understood.  A few lines above there is
```
q_up.prepare("UPDATE config SET cvalue=:val WHERE ckey=:key", qf::core::Exception::Throw);
...
q_ins.prepare("INSERT INTO config (ckey, cvalue, ctype) VALUES (:key, :val, :type)", qf::core::Exception::Throw);
```
Should these occurrences of exception throwing be replaced with return value checking too?

fvacek: Maybe try-catch around all the sql calls will be better. I agree that sql calls should be handled consistently

lpechacek:
take 2 - Patches updated to include feedback.

Issue #233, opened 2016-12-20
No logging data in Application log window


lpechacek:
With latest QE (origin/master as of today) there is no messages in the application log window. I have bisected the bug down to commit 082a6a21cf4c2d59a4e721cd0fd45c9d0af21 (default log category changed...).

lpechacek:
Fix merged into master.

PR #234, opened 2016-12-21
Fix logging into application window (issue #233)

lpechacek:
Commit 082a6a21cf4c changed the default category name.  The same name must be
known to the application log window.


fvacek:
Thanks :)

PR #397, opened 2019-05-14
Some strings are no longer translatable + typos

paukert:
if possible, update TS files in order to finish czech localization, thanks


fvacek:
dobra prace :)

PR #416, opened 2019-07-31
EmmaClient (RaCom type) + Bibs in startlists 

arnost00:
EmmaClient service - RaCom type export full functional for normal race, n-stages race. Experimental for relay.
Bibs - bibs in startlist print (classes, clubs, starters, n-stages events). Better dlg for n-stages event startlist.


fvacek:
docela pekny, dik :)

PR #470, opened 2019-10-16
Print results improved (dialog & wide format)

arnost00:
* Print result for speaker (wide format with club names)
* "First n competitors" moved to result report dialog, available in every result print (normal & multi stage event)


fvacek:
diky

PR #528, opened 2020-03-08
Issue #253: Import start name from ocad import and add it to class start list

xkenia:
Issue #253: Import start name from ocad import and add it to class start list


fvacek:
hele asi se ti do toho commitu dostalo neco jinyho, mam tam porad card checker free order

PR #535, opened 2020-03-16
Comp diag tweaks

otahirs:
* disable buttons and step on edSiId
* remove unnecessary checkbox from Edit Competitor dialog

před
Screenshot from 2020-03-17 19-06-15

po
Screenshot from 2020-03-16 13-05-02

### DIFF
--- a/quickevent/app/plugins/Competitors/src/competitorwidget.ui
+++ b/quickevent/app/plugins/Competitors/src/competitorwidget.ui
@@ -90,6 +90,12 @@
         <property name="alignment">
          <set>Qt::AlignRight|Qt::AlignTrailing|Qt::AlignVCenter</set>
         </property>
+        <property name="buttonSymbols">
+         <enum>QAbstractSpinBox::NoButtons</enum>
+        </property>
+        <property name="singleStep">
+         <number>0</number>
+        </property>
         <property name="dataId" stdset="0">
          <string notr="true">competitors.siId</string>
         </property>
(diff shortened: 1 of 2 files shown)


quickevent/app/plugins/Competitors/src/competitorwidget.ui
```diff
@@ -90,6 +90,12 @@
         <property name="alignment">
          <set>Qt::AlignRight|Qt::AlignTrailing|Qt::AlignVCenter</set>
         </property>
+        <property name="buttonSymbols">
+         <enum>QAbstractSpinBox::NoButtons</enum>
+        </property>
+        <property name="singleStep">
+         <number>0</number>
```
fvacek: why is single step disabled?

otahirs: When entering the SI number, I suppose nobody use the single step functionality to edit that field, therefore I disabled the single step to avoid user confusion or errors. 

At first I wanted to transform spinbox into a line edit box, but in the process I encountered numerous dificulties. In the end I found out it is easier to keep the spinbox and just disable the buttons and single step.

fvacek: SpinBox with arrow doing nothing is much more confusing than one with +1 func as usual.

fvacek:
There is class `SiIdEdit` already, designed to edit SI in forms. If you want to make this class more user friendly, please create an issue with proposed change. We can improve (current poor implementation) this way.

quickevent/app/plugins/Competitors/src/competitorwidget.ui
otahirs: no arrows are displaying thanks to `QAbstractSpinBox::NoButtons`

otahirs:
ok

Issue #66, opened 2015-09-23
Numbers only edit boxes support


Amaroun:
It's a feature request.
If qe is used on ntb without numerical keyboard, switching between CZ/EN keyboards or using shift to enter numbers is very time consuming.
Allowing to configure sort of 'virtual numeric keyboard' for editboxes for numbers edit only, to translate czech characters to coresponding numbers on english keyboard would be of a big help (si number is most common case).

fvacek:
How we will localize such a feature?

otahirs:
Localization is not a problem. Every key on the keyboard has also a static numerical key code.
https://www.w3.org/2002/09/tests/keys.html

fvacek:
Just to be 100% sure. This feature will cause that when one enters `+ěščřžýáíé` in SI number edit box, then entered value will be `1234567890`. Correct?

otahirs:
I tried to implement this. I have not found a way to get to right key code in Qt. None of the methods of [QKeyEvent](https://doc.qt.io/qt-5/qkeyevent.html) shows the original pressed key for Czech keyboard. So localization is a problem after all.

fvacek:
You can use https://doc.qt.io/qt-5/qkeyevent.html#nativeScanCode, for example key `1` has always `native scan code == 10` regardless of keyboard layout used.

otahirs:
I thought so, however `native scan code` in my system for key 1 showed as up to be `32767`

fvacek:
> I thought so, however `native scan code` in my system for key 1 showed up to be `32767`

can you double check this? What is your system? My is Debian 10 x64

otahirs:
I just did that to write the "actual" code on my system. Fedora 32 x64
keydown/keyapp in this webapp works https://www.w3.org/2002/09/tests/keys.html for me, but I had no luck finding anything similar while playing with the `keyPressEvent` event and corresponding `QKeyEvent` class

fvacek:
please try this, I do not see any reason, why it should work differently on Debian. 
[keycodetest.tar.gz](https://github.com/Quick-Event/quickbox/files/5147252/keycodetest.tar.gz)

otahirs:
ok, you minimal working example works great, the native key code is correct (e.g. `10` for key 1)
I might have made a mistake while using the debugger. On the third check, the native scan code is correct even in QE. 
...
Is there a list of constants for native scan codes?
[...]

PR #562, opened 2020-08-31
make spinbox number input independant of keyboard locale #66

otahirs:
need testing on windows @xkenia 
#66

### DIFF
--- a/libqf/libqfqmlwidgets/src/spinbox.cpp
+++ b/libqf/libqfqmlwidgets/src/spinbox.cpp
@@ -1,5 +1,5 @@
 #include "spinbox.h"
-
+#include <QKeyEvent>
 using namespace qf::qmlwidgets;
 
 SpinBox::SpinBox(QWidget *parent)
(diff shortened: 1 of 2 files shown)


paukert:
If I understand correctly, what this should do, it is NOT working for me.

I assume that this should allow me to write numbers while pressing keys +ěščřžýáíé in SI fields, for example here:

OS: W10, version: 2004, build: 19041.450
Built with Qt 5.15.0

si

otahirs:
in windows the native key codes goes from `2` to `11` 
instead of from `10` to `19` on linux

paukert:
Yes, I can confirm that it is working fine with 2-11 on windows.

otahirs:
I did some research.
The range `2-11` is standard both for Windows and Linux.
The commonly used window server for Linux `Xorg` adds `8` to the key code due to historical reasons. 
Wayland (upcoming replacement for Xorg on Linux platform) removes this offset.

In conclusion, I would suggest removing the offset on linux platform for now, as Qt runs under X compatibility layer on Wayland anyway, but keep an eye on this, because if Qt switch to Wayland in the future, we don't know if they will keep the offset for backwards compatibility or if it will be removed.

libqf/libqfqmlwidgets/src/spinbox.cpp
```diff
@@ -34,3 +34,32 @@ void SpinBox::onValueChanged(int n)
...
+		case 10:
+			text = QString::number(code - 1);
+			break;
+		case 11:
+			text = "0";
+			break;
+		default:
+			text = event->text();
+	}
+	QKeyEvent * e = new QKeyEvent(event->type(), event->key(), event->modifiers(), text, event->isAutoRepeat(), event->count());
```
fvacek, line 63: I think that this is a memory leak, you should IMO set value of spinbox here instead of creating fake event

fvacek (changes requested):

libqf/libqfqmlwidgets/src/spinbox.cpp
otahirs: not faking the event would require non trivial reimplementation of the Qt functionality. I admit there might be a memory leak though.

PR #563, opened 2020-08-31
allow clearing spinbox when editing SI card number

otahirs:
Currently, when SI card number edit box is cleared and you leave the edit box, the SI card number appears again. It is not possible to just clear the editing field unless the card number is set explicitly to 0.

this commit makes it so when SI card number is cleared, it is set to 0 instead.

code inspiration: https://codereview.qt-project.org/c/qt/qtbase/+/121767


paukert:
It is working fine in case 1. In second (third) case I have to press enter to confirm deletion (but it is also improvement from previous state). Other fields in Runs tab (start, time) do not requiere confirmation by pressing enter.

example

example2

PR #575, opened 2020-09-16
github actions | build for linux and windows

otahirs:
https://github.com/Quick-Event/quickbox/issues/564#issuecomment-693526488

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -1,5 +1,5 @@
-name: C/C++ CI
 
+name: Build Test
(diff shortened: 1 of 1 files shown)


fvacek:
If this will work, than it will be great !!

fvacek:
Do you think, that it might be able to create artifacts like 
* quickevent-linux64-x.y.x.tgz
* quickevent-linux64-x.y.x.Appimage
* quickevent-win64-x.y.x-setup.exe

otahirs:
I will look into it

PR #577, opened 2020-09-20
fix assigncard autoprint receipt

otahirs:
It might be a good idea to print the receipt after the name is actually updated in the database :sweat_smile:


fvacek:
:)

PR #602, opened 2020-11-09
Add Q_OBJECT macro

paukert:
Added Q_OBJECT macro to make `tr` function work, improved english localization (mentioned in #331).

### DIFF
--- a/quickevent/app/plugins/Classes/src/classeswidget.cpp
+++ b/quickevent/app/plugins/Classes/src/classeswidget.cpp
@@ -106,7 +106,7 @@ private:
 
 class CourseCodesTableModel : public qfm::SqlTableModel
 {
-	Q_DECLARE_TR_FUNCTIONS(CourseCodesTableModel)
+	Q_OBJECT
 private:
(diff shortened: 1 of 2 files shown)


quickevent/app/plugins/Classes/src/classeswidget.cpp
```diff
@@ -106,6 +106,8 @@ class CourseItemDelegate : public QStyledItemDelegate
 
 class CourseCodesTableModel : public qfm::SqlTableModel
 {
+	Q_DECLARE_TR_FUNCTIONS(CourseCodesTableModel)
+private:
```
dg0yt, line 110: This class is a descendant of `QObject`. The canonical way to support `tr` is the `Q_OBJECT` macro. `Q_DECLARE_TR_FUNCTIONS` is for other classes.

paukert: Thanks, it is fixed.

fvacek:
@dg0yt thanks for review

PR #608, opened 2021-04-12
fix deprecated implicit copy assignment operator

otahirs:
```
warning: implicitly-declared ‘qf::core::String& qf::core::String::operator=(const qf::core::String&)’ is deprecated [-Wdeprecated-copy]
```

### DIFF
--- a/libqf/libqfcore/src/core/string.h
+++ b/libqf/libqfcore/src/core/string.h
@@ -17,6 +17,11 @@ public:
 	String() : QString() {}
 	String(const QString &str) : QString(str) {}
 	String(const String &str) : QString(str) {}
+	qf::core::String& operator=(const String &str) {
+		clear();
+		append(str);
+		return *this;
+	}
 public:
 	static const bool TrimParts = true;
 public:
--- a/libqf/libqfcore/src/core/string.h
+++ b/libqf/libqfcore/src/core/string.h
@@ -16,12 +16,6 @@ class QFCORE_DECL_EXPORT String : public QString
 public:
 	String() : QString() {}
 	String(const QString &str) : QString(str) {}
-	String(const String &str) : QString(str) {}
-	qf::core::String& operator=(const String &str) {
-		clear();
-		append(str);
-		return *this;
-	}
 public:
 	static const bool TrimParts = true;
 public:


libqf/libqfcore/src/core/string.h
```diff
@@ -17,6 +17,11 @@ class QFCORE_DECL_EXPORT String : public QString
 	String() : QString() {}
 	String(const QString &str) : QString(str) {}
 	String(const String &str) : QString(str) {}
+	qf::core::String& operator=(const String &str) {
```
fvacek, line 20: this is base class implementation
```c++
String &QString::operator=(const QString &other) noexcept
{
    other.d->ref.ref();
    if (!d->ref.deref())
        Data::deallocate(d);
    d = other.d;
    return *this;
}
```
what is the gain of proposed change? Why throw away QString implicit sharing zero copy?

fvacek (changes requested):

libqf/libqfcore/src/core/string.h
otahirs: compiler throws warnings because `The generation of the implicitly-defined copy assignment operator is deprecated if T has user-declared copy constructor. (since C++11)`  

this change introduces explicit copy assignment operator, so the build log is not cluttered with the warnings mentioned 

the implementation is not *nice* because the QString class is not meant to be inherited in the first place so only public functions are being used. there might be better way to implement this

dg0yt: Rather than introducing another explicit definition, you probably only need to *explicitly* default this operation to resolve this warning:
~~~
String& operator=(const String &str) = default;
~~~
In this class, this recommendation would proably apply to all special member function. Or just remove them all (rule of zero).

fvacek: exactly

fvacek:
```c++
	String() : QString() {}
	String(const QString &str) : QString(str) {}
```
previous explicit ctors decl might be IMO removed as well

dg0yt: Rule of zero probably doesn't work because there is one custom constructor (from QString). So the price for this convenience class (adding members to QString) is explicit (defaulted) special member functions. 
(The alternative is to stick with plain QString, and use free functions.)

otahirs: As I understand it, this should be the minimal implementation.
```cpp
String() = default;
String(const QString &str) : QString(str) {}
```

PR #618, opened 2021-06-10
fix missing refference warning

otahirs:


fvacek:
Diky, ze ses do tech issues pustil :+1:

PR #624, opened 2021-06-13
refactor QDateTime to ISO 8601 with offset string

otahirs:


libquickevent/libquickeventcore/src/utils.cpp
```diff
@@ -5,26 +5,10 @@
...
-				offset_min = -offset_min;
-			}
-			else {
-				ret += '+';
-			}
-			ret += QStringLiteral("%1:%2").arg(offset_min / (60 * 60), 2, 10, QChar('0')).arg((offset_min / 60) % 60, 2, 10, QChar('0'));
-		}
-	}
-	return ret;
+	dt.setOffsetFromUtc(dt.offsetFromUtc());
```
fvacek, line 10: neni toto idempotence?

otahirs: Its not a bug, its a feature. :)

https://bugreports.qt.io/browse/QTBUG-26161?focusedCommentId=554227&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-554227

Tested it does not work - the time zone offset is missing - without this line. (at least as for today).

fvacek: interesting :)

fvacek: please add comment with short explanation of this strange line.

PR #627, opened 2021-06-14
simplify EditOnPunch

### LINKED ISSUE #492: zjednodušení Reader mode | edit on punch
Předpokládá, že reader mode _Entries_ je pouze pro funkcionalitu _edit on punch_.
V tomto připadě bych tedy checkbox _edit on punch_ úplně zrušil a reader mode by pak mohl mít dvě možnosti jako doteď:

-  _Readout_ 
-  _Edit runner on punch_ 

Samotný checkbox mi již přijde nadbytečný.

otahirs:
resolve  #492
purge `Race` and `Entries` markings, remove edit on punch competitor checkbox
introduce `Readout` and `Edit on punch` reader modes.

edit on punch dialog now shows only on connected computer, no db events needed

I am not sure if I the function calls are made as simple as possible, I am not that used to Q_SLOTS/Q_SIGNALS.
Maybe dbversion should change, due to *markigs* being removed.


quickevent/app/plugins/CardReader/src/cardreaderwidget.h
```diff
@@ -52,6 +52,7 @@ class CardReaderWidget : public QFrame
 	Q_OBJECT
 private:
 	typedef QFrame Super;
+	enum PunchMode { Readout, Edit_on_punch };
```
fvacek, line 55: `EditOnPunch`

quickevent/app/plugins/Competitors/src/competitorswidget.h
```diff
@@ -40,18 +40,16 @@ class CompetitorsWidget : public QFrame
...
-
-	void onDbEventNotify(const QString &domain, int connection_id, const QVariant &data);
-	void editCompetitorOnPunch(int siid);
+	Q_SLOT void editCompetitorOnPunch(int siid);
 
 	void onCustomContextMenuRequest(const QPoint &pos);
 
 	void report_competitorsStatistics();
 private:
+	bool editCompetitorLock = false;
```
fvacek, line 49: `m_editCompetitorLock`, but I do not like locks as member fields. They are left unlocked in case of exception and app stop to work. Try to make lock-free implementation or use some kind of scope guard to avoid stay-locked-on-exception

fvacek (changes requested):

quickevent/app/plugins/Competitors/src/competitorswidget.h
otahirs: ps: previous implementation had exactly this functionality, but I will look into it.

fvacek: but checkbox can be explicitly unlocked by user because its state is visible and controllable in UI

quickevent/app/plugins/Competitors/src/competitorswidget.cpp
```diff
@@ -217,43 +211,49 @@ void CompetitorsWidget::editCompetitor_helper(const QVariant &id, int mode, int
...
-	}
-	dlg.setCentralWidget(w);
-	w->load(id, mode);
-	auto *doc = qobject_cast<Competitors::CompetitorDocument*>(w->dataController()->document());
-	QF_ASSERT(doc != nullptr, "Document is null!", return);
-	if(mode == qfm::DataDocument::ModeInsert) {
-		if(siid == 0) {
-			int class_id = m_cbxClasses->currentData().toInt();
-			doc->setValue("competitors.classId", class_id);
+	if(auto lock = std::unique_lock<std::mutex>{m_editCompetitorMutex, std::try_to_lock})
```
fvacek, line 214: the mutex is not necessary, what about
```c++
class EditGuard
{
public:
    EditGuard(bool &lock) : m_lock(lock) { m_lock = true; }
    ~EditGuard() { m_lock = false; }
private:
    bool &m_lock;
}
if(m_insideEditCompetitor) {
    qfDebug() << "Another competitor dialog is opened, ignoring..";
}
else {
    EditGuard g(m_insideEditCompetitor);
   ...
}
```

quickevent/app/plugins/Competitors/src/competitorswidget.h
otahirs: truth, anyway solved using RAII mutex

quickevent/app/plugins/Competitors/src/competitorswidget.cpp
otahirs: why creating explicit class, when mutex is sufficient and designed for locking in the first place?

fvacek: Because purpose of Mutex is to synchronize threads. Using mutex in single-thread app can be confusing for contributors.

fvacek: ScopeGuard is c++ idiom for cases like that https://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/Scope_Guard

otahirs: The Edit dialog can be trigerred by EditOnPunch, I thought there could be multiple threads in play, however I am not sure, how Qt application handles that events. If that is not a concern, its not needed, but I would still prefer mutex over additional class, just because its less code to maintain.

Also lock_guard or scoped_lock could be used for more clarity, but the need for manual unlocking lead me for unique_lock to avoid additional nesting.

arnost00: I think it's better to use Mutex than to write an explicit class. And we are ready for the future (which is multi-threaded) .

fvacek: > The Edit dialog can be trigerred by EditOnPunch, I thought there could be multiple threads in play, however I am not sure, how Qt application handles that events. If that is not a concern, its not needed, but I would still prefer mutex over additional class, just because its less code to maintain.

Whole QT UI must be run in single thread by QT design

fvacek: > I think it's better to use Mutex than to write an explicit class. And we are ready for the future (which is multi-threaded) .

not in QtWidgets

otahirs: now that i have cardreader on hand, I can say that no lock (class nor mutex) is needed, as the thread looking for punch is already blocked by the displayed dialog

fvacek: No it is not. QDialog just creates own QEventLoop object, what does mean that signals from timers or sockets or ... can be delivered meanwhile.

otahirs: To be honest I do not fully understand QT machinery yet. However, If you have edit competitor dialog opened, SI card readout dos not work (not even a beep from the box).

fvacek: try to insert this snippet before `dlg.exec()` in edit competitor
```c++
	QTimer tm;
	connect(&tm, &QTimer::timeout, []() {
		qfInfo() << "timeout";
	});
	tm.start(1000);
	bool ok = dlg.exec();
```
you will see "timeout" in log every second whilst dialog will be open

otahirs: ok, noted.
This still does not explain why, If you have edit competitor dialog opened, editcompetitor dialog is not reopened on repeated SI card punch, without any guard introduced. (I am curious about this, not that is an issue or point for necessary discussion)

If we assume a guard is needed anyway, then i still think mutex is a better solution, i would go further and say it is the only right solution as any other without a mutex is prone to race contidion and potentially locking the edit competitor dialog from opening. (That statement assumes all events from the app are not synchronized at some point and handled by one event loop)

fvacek: > This still does not explain why, If you have edit competitor dialog opened, editcompetitor dialog is not reopened on repeated SI card punch, without any guard introduced. (I am curious about this, not that is an issue or point for necessary discussion)

I guess that this is SQL concurrency issue and that this is rather bug. Qt itself allows concurrent SI reading while edit competitor dialog is open.

fvacek: > If we assume a guard is needed anyway, then i still think mutex is a better solution

hijacking mutex in single threaded application is always bad solution, it is also bad pattern for contributors and it makes code understanding harder. Pls, see our discussion above about Qt concurency, the mutex introducion introduces also many questions about rest of code.

fvacek:
conflicts from last merge :((

PR #629, opened 2021-06-18
simplify access to other plugins

otahirs:
using this function template
```cpp
template<typename T>
static T* getPlugin()
{
	static_assert( std::is_base_of<Plugin, T>::value, "given type is not Plugin");
	qf::qmlwidgets::framework::MainWindow *fwk = qf::qmlwidgets::framework::MainWindow::frameWork();
	return fwk->plugin<T*>();
}
```

I would like to remove every bit of code of this kind:
```cpp
static Event::EventPlugin* eventPlugin()
{
	qf::qmlwidgets::framework::MainWindow *fwk = qf::qmlwidgets::framework::MainWindow::frameWork();
	auto *plugin = qobject_cast<Event::EventPlugin*>(fwk->plugin("Event"));
	QF_ASSERT_EX(plugin != nullptr, "Bad Event plugin!");
	return plugin;
}
```
and than use the template as `getPlugin<EventPlugin>()`
see the concept in the commit

eventually what name to use? can tou come up with better name than `getPlugin<>`?


fvacek:
perfektni :+1:

otahirs:
what is the difference between 
```cpp
getPlugin<ReceiptsPlugin>()->printReceipt(card_id);
```
and 
```cpp
QMetaObject::invokeMethod(getPlugin<ReceiptsPlugin>(), "printReceipt", Qt::DirectConnection, Q_ARG(int, card_id));
```

which one should be prefered? when its is neede to *invoke* instead of calling directly?

otahirs:
If anyone have a tip why locally Qt Creator has no problem building this branch on both Linux and Windows, but `make-dist.sh` and `make-win.sh` fail to do so, you are welcome. 

Builds fail on `cannot find -lRunsQEPlugin` etc. added in bd76c65

otahirs:
without Windows it would be possilble to use `getPlugin<FooPlugin>()` anywhere, however Windows does not like linking libraries with cyclic dependencies. (on Linux it was possilbe to just erase all `LIBS += ....` references and *it just works*)
So if there is already a usage of plugin B in plugin A, its necesery to use `invokeMethod` or `singnals/slots` when caling A from B.

otahirs:
I am wondering, since the idea of separate plugins was scrapped, can not we just compile all plugins/whole quickevent as a single library?

fvacek:
> what is the difference between
> 
> ```c++
> getPlugin<ReceiptsPlugin>()->printReceipt(card_id);
> ```
> 
> and
> 
> ```c++
> QMetaObject::invokeMethod(getPlugin<ReceiptsPlugin>(), "printReceipt", Qt::DirectConnection, Q_ARG(int, card_id));
> ```
> 
> which one should be prefered? when its is neede to _invoke_ instead of calling directly?

The firs one is resolved on compile time whilst the second one on the run-time. The first one is the preferred one.

fvacek:
> I am wondering, since the idea of separate plugins was scrapped, can not we just compile all plugins/whole quickevent as a single library?

We can, for sure, we only need one who have spare time to do it. I like to have it this way.

otahirs:
Its still along way to go, not fully without quirks yet, but I managed to build it in one piece.
I think, when finished, it will enable to call functions directly without `invoke` and maybe it will be possible to get rid of some code currently used for plugin initialization.

otahirs:
ready for review

fvacek:
It is pretty huge commit, I'll try to go though it during weekend. Anyway the idea is great and I'm glad that you did it. :+1: May be, this is so big change, that we can also bump app version major number and create dedicated branch for the testing time. Thank you, good job.

otahirs:
oblastní žebříček na této branchi úspěšně uspořádán.
jedinný objevený problém zapomenutého řádku 55f07dc
(+ nesouvisející #632)

quickevent/app/quickevent/plugins/CardReader/CardReader.pri
```diff
@@ -0,0 +1,32 @@
...
+include (src/src.pri)
+
+RESOURCES += \
+      $$PWD/$${PLUGIN_NAME}.qrc
+
+OTHER_FILES += \
+        $$PWD/qml/reports/* \
+
+TRANSLATIONS += \
+        $$PWD/$${PLUGIN_NAME}.cs_CZ.ts \
```
fvacek, line 14: I'm afraid that we have to merge also localization into one file per language somehow

otahirs: this is not needed, as for now we do not use qmake to build the language files.

There is a possibility to move the language file generation into the qmake, but as for now we use a special parts of the build scripts for that.

otahirs: the `TRANSLATIONS` variable has been actually redundant all the time.
In a new issue we can discuss handling of the translations, its possible to simplify it, it will take some time though. using `CONFIG += lrelease embed_translations`, changing the .ts files naming style (multiple dots in the name cause problems), refactor translation loading and possibly merging the files into one,..

fvacek:
It looks good to me. There are some conflicts due other pull requests, please fix them. I still think, that we should switch to QE 2.4.0 with this change.

quickevent/app/quickevent/plugins/Competitors/src/competitorsplugin.cpp
```diff
@@ -72,8 +66,7 @@ int CompetitorsPlugin::editCompetitor(int id, int mode)
 void CompetitorsPlugin::onInstalled()
 {
 	qff::MainWindow *fwk = qff::MainWindow::frameWork();
-	m_partWidget = new ThisPartWidget();
-	fwk->addPartWidget(m_partWidget, manifest()->featureId());
+	qff::initPluginWidget<CompetitorsWidget, PartWidget>(tr("Competitors"), featureId());
```
otahirs, line 69: @fvacek please wait for resolving  #627 before releasing new version
@paukert in this branch are introduced new translated strings in the plugins like this, I think the words are already translated, but the translation files are not updated, please do some magic to fix this.

paukert: Run the following command with correct paths, it will automatically update all TS files:
`QT_DIR/bin/lupdate-pro.exe -no-obsolete quickbox/quickbox.pro`

otahirs: the Translation files are still referenced in the .pro file though.
I have rewritten the strings to match the one in translations files and ran the update command.
The number of additions seems off (I do not know what is the issue here), but it works.

paukert: I went quickly through CardReader.cs_CZ.ts and it contains also translations for Receipts plugin etc. Same problem occurs propably in every TS file :(

otahirs: So I guess we need to merge the files for each language into one?

otahirs:
Thanks to merging the translation files into one I was able to ship the files during the application build 26e4050. 
Additional scripts for copying the files no longer needed.

I needed to rename the language files from `quickevent.cs_CZ.ts` to `quickevent-cs_CZ.ts` because qmake recognized the dot as an end of filename and threw errors of duplicated files. (all quickevent.*.ts were recognized as one file. most likely a bug)

paukert:
Translations look good :) After merging PR it will be necessary to update project on Weblate but that is not important now.

fvacek:
Good job, thanks a lot. :+1:

PR #635, opened 2021-06-28
Add qtbase translations

paukert:
In currently used QT version there are no translations in qt_xxx.qm file, most of them were probably moved to qtbase_xxx.qm

This PR reflects this change and copies these translations into the QE translations folder.

I hope I didn't miss some build script which is still being used.


dg0yt:
For most version of Qt5, the canonical way was to create, deploy and use a single file per language by:
~~~
lconvert -o qt_cs.ts qt_cs.ts qtbase_cs.ts <qtanothermodule_cs.ts...>
~~~

Now in Qt 5.15, they seem to propose loading the needed modules explicitly - which implies *deploying* them all individually.
https://doc.qt.io/qt-5/internationalization.html#available-catalogs

paukert:
So far, I haven't faced missing translations excluding buttons in widgets so IMO (and according description) loading qtbase could be sufficient for QE.

PR #644, opened 2021-08-02
Update README.md && remove lctools

paukert:
- fix few typos, add build badge to README.md
- lctools are no longer used because translations are automatically embedded in the executable during build and for updating TS files it is easier to run ./{Qt_bin_dir}/lupdate-pro -no-obsolete <root .pro file>


fvacek:
perfektni, diky :)

PR #667, opened 2021-10-02
Fix EmmaClient txt export.

arnost00:
Resolve issue #655 
- DNS in finish file
- Rework finish file, now mainly work from `runs`
- Remove old si card checking from `cards`.
- Fix when disqualification is cancelled.
- Change DISQ & MP logic - same as in IOF XML export

### DIFF
--- a/quickevent/app/quickevent/plugins/Event/src/services/emmaclient.cpp
+++ b/quickevent/app/quickevent/plugins/Event/src/services/emmaclient.cpp
@@ -106,7 +106,7 @@ void EmmaClient::exportRadioCodes()
 		//qfInfo() << qb_codes.toString();
 
 		QString class_name = q1.value("classes.name").toString();
-        class_name.remove(" ");
+		class_name.remove(" ");
 		QVector<int> codes;
 		qfs::Query q2;
 		q2.execThrow(qb_codes.toString());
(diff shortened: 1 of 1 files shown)


paukert:
I went quickly through changes and I think there is a problem with printing all runners to finish file even if we do not have information about them e.g. at the beginning all runners will be marked as DNF and DNS, won't they?

arnost00:
> 
> 
> I went quickly through changes and I think there is a problem with printing all runners to finish file even if we do not have information about them e.g. at the beginning all runners will be marked as DNF and DNS, won't they?

It looks like you are correct. Maybe I need to add back `cards` in select. 

Does anyone have a backup from the beginning, from the middle and after the race? 
I have only backups after race from last month.

quickevent/app/quickevent/plugins/Event/src/services/emmaclient.cpp
```diff
@@ -267,48 +267,54 @@ void EmmaClient::exportFinish()
...
-            .orderBy("cards.id ASC");
 	int start00 = getPlugin<EventPlugin>()->stageStartMsec(current_stage);
 
+	qfs::QueryBuilder qb;
 	qfs::Query q2;
+
+	qb.select2("runs", "siId, isRunning, finishTimeMs, misPunch, badCheck, disqualified, notCompeting")
+			.from("runs")
+			.where("runs.stageId=" QF_IARG(current_stage))
+			.where("runs.siId IS NOT NULL")
```
paukert, line 278: Wasn't be sufficient to filter out runners who do not have check time and are running?

Something like:
`.where("runs.checkTimeMs IS NOT NULL OR runs.isRunning IS NULL")`
or
`.where("runs.checkTimeMs IS NOT NULL OR NOT runs.isRunning")`
(I don't know which type is `isRunning` in DB)

paukert:
During tests I faced strange behaviour:

```sql
SELECT * FROM runs WHERE NOT isRunning
SELECT * FROM runs WHERE isRunning IS NULL
```

results were different accross events, mostly only one option worked.

Maybe we should force notNull in DbSchema
https://github.com/Quick-Event/quickbox/blob/551548ad5e3791f2d398d45511ed13b70a3f2b13/quickevent/app/quickevent/plugins/Event/qml/DbSchema.qml#L208

quickevent/app/quickevent/plugins/Event/src/services/emmaclient.cpp
arnost00: I cannot filter out `isRunning`, because of DNS info.
Better than `checkTimeMs` si check if there is record in `cards`.

arnost00:
It now uses `cards` for check if runner is not in forest.
Runner is now in export if : 
 - is not `isRunning`
 - if has record in `cards`

PR #684, opened 2021-12-06
transform SiCard value from 0 to Empty in edit competitor widget

otahirs:
before
Screenshot from 2021-12-06 11-32-55

after
Screenshot from 2021-12-06 11-31-32

### DIFF
--- a/libquickevent/libquickeventgui/src/si/siidedit.cpp
+++ b/libquickevent/libquickeventgui/src/si/siidedit.cpp
@@ -13,9 +13,23 @@ SiIdEdit::SiIdEdit(QWidget *parent)
 
 void SiIdEdit::fixup(QString &input) const
 {
-    if (input.isEmpty())
-	input = QString::fromLatin1("0");
-    Super::fixup(input);
+	if (input.isEmpty())
+		input = QString::fromLatin1("0");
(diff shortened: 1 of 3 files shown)


arnost00:
Why is 0 value problem ?
And why only in this one dialog ?
In other cases we show 0. In relay runners edit, tableviews, reports and more.

otahirs:
> Why is 0 value problem ? 
it needs to be manually deleted every time

> why only in this one dialog ?
I have not run into inconveniences enywhere else, I also did not look how the value is stored and moved in other places so in order not to break anything I only updated the look in edit field.

arnost00:
> I have not run into inconveniences enywhere else, I also did not look how the value is stored and moved in other places so in order not to break anything I only updated the look in edit field.

Relays? [Add unregistred runner](https://github.com/Quick-Event/quickbox/issues/665#issuecomment-955700517)

otahirs:
form was not using SiIDEdit class, but a simple QSpinBox. I changed it to SiIdEdit as this is used in the edit competitor dialog and now the appearance is the same.

fvacek:
it seems good to me, IMO it doesn't matter if an invalid SI is `0` or empty string as well.

PR #696, opened 2022-01-23
Fix mass start time overwrite by readout of previous leg

otahirs:
recently i added this line to set start times for next legs in advance
```cpp
		/// set start time for next leg
		setStartTimeForNextLeg(relay_id, leg, checked_card.finishTimeMs());
```
https://github.com/otahirs/quickbox/blob/dd567e5f6b358e4da30f0cffd529713d8f1c7045/quickevent/app/quickevent/plugins/CardReader/src/cardreaderplugin.cpp#L502-L503

---
so this check could be implemented
```cpp
		if(is_relays) {
			int run_leg = q.value("leg").toInt();
			if(run_start == 0 && run_leg != 1) {
				/// start time not set => this leg does not event start
				qfInfo() << tr("skipping assign of SI: %1 to run_id: %2; start time not set, this leg does not event start").arg(si_id).arg(run_id);
				continue;
			}
		}
```
https://github.com/otahirs/quickbox/blob/dd567e5f6b358e4da30f0cffd529713d8f1c7045/quickevent/app/quickevent/plugins/CardReader/src/cardreaderplugin.cpp#L124-L131

---
however, setting start time for next legs in advance during readout is not desired if the start time is set manually for runners starting from a mass start, as they would be overwritten.

so to fix the unwanted overwrite of start times I propose deleting both snippets. 
ps: as a sideeffect to this fix, if the same SI Card is used during relays twice, manual assign is needed for both runners as it is impossible to guess if the second runner does the readout or if the first one is reading out again with changes for whatever reason

### DIFF
--- a/quickevent/app/quickevent/plugins/CardReader/src/cardreaderplugin.cpp
+++ b/quickevent/app/quickevent/plugins/CardReader/src/cardreaderplugin.cpp
@@ -444,13 +444,12 @@ void CardReaderPlugin::assignCardToRun(int card_id, int run_id)
 	processCardToRunAssignment(card_id, run_id);
 }
 
-void CardReaderPlugin::setStartTimeForNextLeg(int relay_id, int prev_leg, int prev_finish_time) {
+void CardReaderPlugin::setStartTime(int relay_id, int leg, int start_time) {
 	qf::core::sql::Query q;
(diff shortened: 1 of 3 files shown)


fvacek:
please increase QE version in `quickbox/quickevent/app/quickevent/src/appversion.h` to make potential issue tracking easier

PR #727, opened 2022-04-17
Issue #724 - added more information in XML3.0

arnost00:
Modified XML export -added more information in XML3.0 export (Organisation\Name, Result\TimeBehind, Result\ControlCard)


fvacek:
please increase QE version in https://github.com/Quick-Event/quickbox/blob/master/quickevent/app/quickevent/src/appversion.h to keep track of implemented issues

PR #748, opened 2022-06-09
Draw tool show start time 

arnost00:
Solve Issue #740 - use start time from db


fvacek:
tohle nicemu nepomuze, protoze draw tool ostatni kategorie stejne sesklada za sebe, takze vubec nevim, k cemu by tato uprava mohla prospet

PR #790, opened 2022-08-08
feat: OResults service

otahirs:
Together with @Jurda5, we are developing a new live results service similar to liveresultat.orientering.se, but hopefully more friendly for both users and organizers, also including tight integration with the Bluebox radio controls.  

That being said, we need a new service in QuickEvent for uploading the finish times and result status for runners, this MR should cover that.

### DIFF
--- a/quickevent/app/quickevent/plugins/Event/src/services/oresultsclient.cpp
+++ b/quickevent/app/quickevent/plugins/Event/src/services/oresultsclient.cpp
@@ -32,8 +32,6 @@ using Runs::RunsPlugin;
 namespace Event {
 namespace services {
 
-const QString API_URL = "https://api.oresults.eu";
-
 OResultsClient::OResultsClient(QObject *parent)
 	: Super(OResultsClient::serviceName(), parent)
 {
--- a/quickevent/app/quickevent/plugins/Event/src/services/oresultsclient.cpp
+++ b/quickevent/app/quickevent/plugins/Event/src/services/oresultsclient.cpp
@@ -42,10 +42,8 @@ OResultsClient::OResultsClient(QObject *parent)
 	connect(m_exportTimer, &QTimer::timeout, this, &OResultsClient::onExportTimerTimeOut);
 	connect(this, &OResultsClient::statusChanged, [this](Status status) {
 		if(status == Status::Running) {
-			if(settings().exportIntervalSec() > 0) {
-				onExportTimerTimeOut();
-				m_exportTimer->start();
-			}
+			onExportTimerTimeOut();
+			m_exportTimer->start();
 		}
 		else {
 			m_exportTimer->stop();
@@ -94,14 +92,8 @@ qf::qmlwidgets::framework::DialogWidget *OResultsClient::createDetailWidget()
 void OResultsClient::init()
 {
 	OResultsClientSettings ss = settings();
-	if(ss.exportIntervalSec() > 0) {
-		m_exportTimer->setInterval(ss.exportIntervalSec() * 1000);
-	}
-	else {
-		m_exportTimer->stop();
-	}
+	m_exportTimer->setInterval(ss.exportIntervalSec() * 1000);
 }
-
 
 void OResultsClient::onExportTimerTimeOut()
 {
--- a/quickevent/app/quickevent/plugins/Event/src/services/oresultsclient.cpp
+++ b/quickevent/app/quickevent/plugins/Event/src/services/oresultsclient.cpp
@@ -1,0 +1,155 @@
+#include "oresultsclient.h"
+#include "oresultsclientwidget.h"
+
+#include "../eventplugin.h"
+
+#include <qf/qmlwidgets/framework/mainwindow.h>
+#include <qf/qmlwidgets/dialogs/dialog.h>
+
+#include <qf/core/log.h>
+#include <plugins/Runs/src/runsplugin.h>
+#include <plugins/Relays/src/relaysplugin.h>
+
+#include <QDir>
+#include <QFile>
+#include <QHttpPart>
+#include <QNetworkAccessManager>
+#include <QNetworkReply>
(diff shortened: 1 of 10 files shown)


quickevent/app/quickevent/plugins/Event/src/services/oresultsclient.h
```diff
@@ -0,0 +1,53 @@
...
+
+class OResultsClient : public Service
+{
+	Q_OBJECT
+
+	using Super = Service;
+public:
+	OResultsClient(QObject *parent);
+
+	//void run() override;
```
fvacek, line 33: IMO, `run()` & `stop()` must start/stop export timer, current service implementation cannot be stop, unless one set interval to zero.

fvacek (changes requested):

quickevent/app/quickevent/plugins/Event/src/services/oresultsclient.h
otahirs: starting and stopping the service works fine.
However, i found out another bug. `Ok button` does not save settings, will fix shortly

quickevent/app/quickevent/plugins/Event/src/services/oresultsclient.h
```diff
@@ -0,0 +1,53 @@
+#ifndef ORESULTSCLIENT_H
+#define ORESULTSCLIENT_H
+
+#pragma once
+
+#include "service.h"
+
+#include <QNetworkAccessManager>
```
arnost00, line 8: Must be include here ? Cannot be used like QTimer ?
in header only : ``class QNetworkAccessManager;``

otahirs: It makes no difference. As i see it, `includes` and `class declarations` are used interchangeably in the project. If one is the preferred way, i can change it. It is up to @fvacek

fvacek: I always prefer forward declaration in header file, because it saves compile time.

quickevent/app/quickevent/plugins/Event/src/services/oresultsclient.cpp
```diff
@@ -0,0 +1,155 @@
...
+			if(settings().exportIntervalSec() > 0) {
+				onExportTimerTimeOut();
+				m_exportTimer->start();
+			}
+		}
+		else {
+			m_exportTimer->stop();
+		}
+	});
+	connect(this, &OResultsClient::settingsChanged, this, &OResultsClient::init, Qt::QueuedConnection);
```
fvacek, line 54: The user must change settings to stop service in this implementation.

otahirs: I removed some redundant code. User can stop/start service from the `Services` widget. The running status is not dependent on the exportTimer interval. Saving settings does not interfere with running status.

fvacek: How can user start/stop service, if oresultclient does not implement 
```
 void run() override;
void stop() override;
```
?

otahirs: emmaclient does not implement them either
https://github.com/Quick-Event/quickbox/blob/83c0dc496abc90df879d2465df4bff3f83657d05/quickevent/app/quickevent/plugins/Event/src/services/emmaclient.h#L33-L34

fvacek: Yes, emmaclient should be fixed as well. The service can be stopped/run only by setting export time interval to 0, what is somewhat non-intuitive. I've missed it during emmaclient review. If I new this, I'd make stop/run functions pure virtual to push developers to implement Service interface.

arnost00: Emmaclient - Reaction on emit statusChanged is not enought ? run/stop only calls status update

arnost00: This type of run / stop was introduced by @fvacek in commit [QE: emma client iofxml3 export](https://github.com/Quick-Event/quickbox/commit/6c48e14844604844522d041c5f4425425c2c4101)

otahirs: > Yes, emmaclient should be fixed as well. The service can be stopped/run only by setting export time interval to 0, what is somewhat non-intuitive. I've missed it during emmaclient review. If I new this, I'd make stop/run functions pure virtual to push developers to implement Service interface.

I believe you might be incorrect. Try the artifact build. It worked for me fine. https://github.com/Quick-Event/quickbox/actions/runs/2910037225

fvacek: image
Yes, I'm wrong. Anyway, current implementation where `start()/stop()` functions just change the service status, which is checked by some `if` on strange place `OResultsClient::onExportTimerTimeOut()`, is working but not easy readable, as you can see during this review:) Correct implementation should IMO start/stop export timer in `start()/stop()` re-implementations. This also applies to emmaclient.

otahirs: > Correct implementation should IMO start/stop export timer in start()/stop() re-implementations. 

I agree.

otahirs: Set up an issue and assign me to it, I will fix it later.

fvacek: https://github.com/Quick-Event/quickbox/issues/795

PR #800, opened 2022-09-08
Extend service RacomClient - can reads splits from rawsplits txt file

arnost00:
Reads splits and write them to table "punches".
Reads format known as "racom rawsplits".  
Check every split if not duplicate, before write it.
Some information about in #472


quickevent/app/quickevent/plugins/CardReader/src/cardreaderplugin.cpp
```diff
@@ -298,6 +298,18 @@ int CardReaderPlugin::savePunchRecordToSql(const quickevent::core::si::PunchReco
 
 	int code = resolveAltCode(punch.code(), punch.stageid());
 
+	// check if punch isn't duplicate of existing saved punch
```
fvacek, line 301: what is this good for?

quickevent/app/quickevent/plugins/CardReader/src/services/racomclient.cpp
```diff
@@ -132,6 +135,85 @@ void RacomClientSirxdConnection::onReadyRead()
...
+				QTextStream stream(&file);
+
+				QString line;
+				do {
+					line = stream.readLine();
+					text << line;
+				} while(!line.isNull());
+			}
+			file.close();
+	}  catch (QException &e) {
```
fvacek, line 175: which function in try/catch can throw QException?

quickevent/app/quickevent/plugins/CardReader/src/cardreaderplugin.cpp
arnost00: It checks for duplicity in punches table.
Needed in case of : 
* new read `rawsplits` file (for example after stop/start service)
* second card redout with same finish time

quickevent/app/quickevent/plugins/CardReader/src/services/racomclient.cpp
arnost00: Not sure now which function exactly.
It is there for catch if file cannot be open or read - racom file is on network device (OOk server)

fvacek: QException is not thrown in this case. https://doc.qt.io/qt-5/qexception.html#details

quickevent/app/quickevent/plugins/CardReader/src/services/racomclient.h
```diff
@@ -20,11 +21,48 @@ class RacomClientSettings : public Event::services::ServiceSettings
...
+	RacomReadSplitFile(QString fileName, int interval, int finishCode, QObject *parent);
+	~RacomReadSplitFile();
+	void readAndProcessFile();
+	void run();
+private:
+	QString m_fileName;
+	int m_interval = 5;
+	int m_finishCode = 2;
+	int m_lastRowCount = 0;
+	QTimer *m_ReadTimer = nullptr;
```
fvacek, line 62: `m_readTimer`

quickevent/app/quickevent/plugins/CardReader/src/services/racomclient.cpp
```diff
@@ -132,6 +135,85 @@ void RacomClientSirxdConnection::onReadyRead()
...
 }
 
+RacomReadSplitFile::RacomReadSplitFile(QString fileName, int interval, int finishCode, QObject *parent)
+	: Super(parent)
+	, m_fileName(fileName)
+	, m_interval(interval)
+	, m_finishCode(finishCode)
+{
+	m_ReadTimer = new QTimer(this);
+	connect(m_ReadTimer, &QTimer::timeout, this, &RacomReadSplitFile::onTimerTimeOut);
```
fvacek, line 145: why not `connect(m_ReadTimer, &QTimer::timeout, this, &RacomReadSplitFile::redAndProcessFile);`

quickevent/app/quickevent/plugins/CardReader/src/cardreaderplugin.cpp
fvacek: You are removing duplicates in `readAndProcessFile()` already, I like to have it all on one place. 

I do not know, why duplicate record in DB mind, but if you need to not-save-duplicate-punches do DB, please add comment with explanation before implementation, in other case, we do not know, what is this code good for. The best would be, if you can do this check in `readAndProcessFile()`. If this is not possible, please do not prepare sql query for such simple task.

quickevent/app/quickevent/plugins/CardReader/src/services/racomclient.h
arnost00: I'll fix it

quickevent/app/quickevent/plugins/CardReader/src/services/racomclient.cpp
arnost00: I'll fix it

quickevent/app/quickevent/plugins/CardReader/src/cardreaderplugin.cpp
arnost00: > You are removing duplicates in `readAndProcessFile()` already, I like to have it all on one place.

In `readAndProcessFile()` it remove duplicate from rawsplits file.
In `savePunchRecordToSql(...)` it checks if record doesn't exist in DB (finish from radio x finish from readout / 2nd readout of card)

> I do not know, why duplicate record in DB mind ...

The reason is, we don't want have duplicity in `punches` table. Because of Speaker module, which uses this table. If we don't check for duplicity on inserting in db, we need to check it on every shown in module Speaker.
I think it is better to resolve duplicity on write, and not on every read.

> ... please add comment with explanation before implementation, in other case, we do not know, what is this code good for. ...

I was hoping comment `check if punch isn't duplicate of existing saved punch` is enough. What information are you missing in this comment? I'm not sure what needs to be added?

arnost00: > If this is not possible, please do not prepare sql query for such simple task.

I was hoping it is more readable with `prepare`.  Can be rewrite to `q.exec("SELECT * ...`

fvacek: > > ... please add comment with explanation before implementation, in other case, we do not know, what is this code good for. ...
> 
> I was hoping comment `check if punch isn't duplicate of existing saved punch` is enough. What information are you missing in this comment? I'm not sure what needs to be added?

somethink like:
```
The reason is, we don't want have duplicity in punches table. Because of Speaker module, which uses this table. If we don't check for duplicity on inserting in db, we need to check it on every shown in module Speaker.
I think it is better to resolve duplicity on write, and not on every read.
```

fvacek: > > If this is not possible, please do not prepare sql query for such simple task.
> 
> I was hoping it is more readable with `prepare`. Can be rewrite to `q.exec("SELECT * ...`

Prepare query is more resource consuming for SQL server, they are useful in case, when one is executing the same query in the loop with different parameters. https://www.postgresql.org/docs/current/sql-prepare.html We are sometimes using prepared queries because `QSqlQuery::bind()` can correctly send QDateTime parameter to server with all the TZ stuff solved.

fvacek: > The reason is, we don't want have duplicity in punches table. Because of Speaker module, which uses this table. If we don't check for duplicity on inserting in db, we need to check it on every shown in module Speaker.
I think it is better to resolve duplicity on write, and not on every read.

This is not always true. 
1. If you want to state, that punch records shall be unique, then SQL unique index is the right way. Than you can have checks on many places or you can catch the SQL exceptions, but you do not have to think, if you would might to forget about duplicity check in some program path.
2. If you allow duplicates, then the better place to fix them is the place which do not need them. So `SELECT DISTINCT` is your friend.

This is something what must be discussed.

arnost00: I'll extend the comment & remove q.prepare.

>This is something what must be discussed.
I was hoping it is good enough. I'm not big expert in SQL. So, I'll wait until you resolve what is the best way to do it in SQL db. 
I just thought that duplicate data in the db are unwanted. They just increase the amount of data in the table and don't carry any information.

quickevent/app/quickevent/plugins/CardReader/src/services/racomclient.cpp
arnost00: I'll try to check it and fix it.

fvacek: any progress?

arnost00: Yes, I've remove try/catch

fvacek:
Last thing, please increment QE version number to have track, which version has implemented this feature and which one has not.

PR #812, opened 2022-10-06
fix: unify time export format

### LINKED ISSUE #808: XML export time format differences
The function used for converting datetimes to string in XML export is not the same:

Individual startlist
`2022-01-22T12:00:00.000Z`

Individial resutls
Relay startlist
Relay results
`2022-01-22T13:00:00+01:00`

Both formats are technically correct, but we should unify the export format.
The version with time zone makes more sense to me, to do not lose time zone information.

otahirs:
close #808


fvacek:
why is the necrolog included to this PR?

fvacek (changes requested):
remove necrolog from PR

PR #815, opened 2022-10-10
refactor: improve run flags dialog

### LINKED ISSUE #714: Add disqualified explicitly option to run flags
image

### LINKED ISSUE #796: Run flags improvements
1. whats the diference bewtween the first two options?
image

---
2. the top check box `Disqualified` gets confusing really quickly. It gets selected automatically, but also can be un-checked afterwards.

otahirs:
resolve  #796, close #714
status is displayed as one of `OK`, `Disqualified`, `Not competing`
Screenshot from 2022-10-10 13-17-20

### DIFF
--- a/quickevent/app/quickevent/plugins/Runs/src/runflagsdialog.h
+++ b/quickevent/app/quickevent/plugins/Runs/src/runflagsdialog.h
@@ -26,6 +26,8 @@ private:
 	Ui::RunFlagsDialog *ui;
 	RunsTableModel *m_model;
 	int m_row;
+	bool m_isDisqualified;
+	void updateStatusText();
 };
 
 
--- a/quickevent/app/quickevent/plugins/Runs/src/runflagsdialog.h
+++ b/quickevent/app/quickevent/plugins/Runs/src/runflagsdialog.h
@@ -22,12 +22,12 @@ public:
 	void save();
 private:
 	void updateState();
+	void updateStatusText();
 private:
 	Ui::RunFlagsDialog *ui;
 	RunsTableModel *m_model;
 	int m_row;
 	bool m_isDisqualified;
-	void updateStatusText();
 };
 
 
--- a/quickevent/app/quickevent/plugins/Runs/src/runflagsdialog.h
+++ b/quickevent/app/quickevent/plugins/Runs/src/runflagsdialog.h
@@ -21,8 +21,8 @@ public:
 	void load(RunsTableModel *model, int row);
 	void save();
 private:
-	void updateState();
-	void updateStatusText();
+	void updateStatus();
+	bool isDisqualified() const;
 private:
 	Ui::RunFlagsDialog *ui;
 	RunsTableModel *m_model;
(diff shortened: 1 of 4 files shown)


quickevent/app/quickevent/plugins/Runs/src/runstablemodel.cpp
```diff
@@ -68,27 +68,27 @@ QVariant RunsTableModel::value(int row_ix, int column_ix) const
...
 		if(not_start)
-			sl << tr("NS", "DidNotStart");
+			sl << tr("DNS", "DidNotStart");
 		if(not_finish)
-			sl << tr("NF", "DidNotFinish");
+			sl << tr("DNF", "DidNotFinish");
+		if(is_disqualified_by_organizer)
+			sl << tr("DO", "disqualifiedByOrganizer");
+		if(is_disqualified && !mis_punch && !bad_check && !not_start && !not_finish && !is_disqualified_by_organizer)
+			sl << tr("DSQ", "Disqualified");
```
otahirs, line 91: `DSQ` is displayed only if other flags are missing to provide more clarity to the user. (shows only `MP` instead of `DSQ, MP`)

quickevent/app/quickevent/plugins/Runs/src/runflagsdialog.h
```diff
@@ -26,6 +26,8 @@ class RunFlagsDialog : public QDialog
 	Ui::RunFlagsDialog *ui;
 	RunsTableModel *m_model;
 	int m_row;
+	bool m_isDisqualified;
```
fvacek, line 29: no need for new class field, since `m_isDisqualified` can be computed anytime from checkboxes

quickevent/app/quickevent/plugins/Runs/src/runflagsdialog.h
```diff
@@ -26,6 +26,8 @@ class RunFlagsDialog : public QDialog
...
+	void updateStatusText();
```
fvacek, line 30: please don't mix fields and methods declarations

fvacek (changes requested):

quickevent/app/quickevent/plugins/Runs/src/runflagsdialog.h
otahirs: I was always wondering, why there are multiple `private` and `public` blocks in header files. now i do understand :smile:

fvacek: This is not explicitly mentioned in code style, but we are using Qt code style, if explicit rule is not stated.

quickevent/app/quickevent/plugins/Runs/src/runflagsdialog.h
otahirs: as there is a need to sync the state with the db entry  
i would argue this approach is more safe in case of future edits
```cpp
  table_row.setValue(QStringLiteral("runs.disqualified"), m_isDisqualified);
```

fvacek: what about `private bool isDisqualified() const;`?

otahirs: Seems good to me. :handshake:

fvacek: the best class field is no-field :)

Issue #691, opened 2022-01-10
Card Flags improvements


otahirs:
1. use the new `{run|card}flags` columns and dialogs from recent commit 3a3c82b078439da7e7d67c5b4cfcb40e666bb1b2 also in the runs table view in competitor edit dialog.

2. use the `run flags` column in the `Readout module`

3. ~~a) should the change in `card in rent table` flag also add/remove the card to/from rent table?~~
  ~~should this flag be read only~~ 
  ~~c) no change~~
  flag in dialog is already read only (non-clickable), maybe we should give it disabled look and add a tool tip explaining where to edit that
  
4. `Readout module` -> `card in rent table` column should be read only, `card returned` column should be disabled if card is not rented  (or merge these two columns into one?)

fvacek:
There are two ways, how to rent a card.
1. Organizer mark every rented card in run, when the card is passed to competitor
2. Organizer has table of all SI of cards supposed to be rent. In this case, no action on the organizer side is needed during card rent.

We have to keep both IMO.

otahirs:
I corrected **n3**.

to add to **n4**:
the `RT` column has tooltip `card in rent table`, although it has a tick if either  `card is in rent table` or `rent is requested` is true, so the tooltip is misleading. Also as I already stated, this column should be read only as ticking or clearing the checkbox does nothing, if the view is refreshed, this column goes back to previous state.

To expand it further, the naming `Card requested` is misleading as not only this flag means that the runner requested to rent a card, but also that the card has been actually rented out. I would advocate to change the naming from 
```
Card rent requested
Card in rents card table
```
to maybe something like
```
Card rented (rent requested)
Card rented (from rent table)
```
it is up to discussion

paukert:
As far as I know, card rent requested flag is automatically set during import from ORIS if runner asked for SI card. In most cases runner will come for SI card before race but theoretically he could also borrow SI card from someone else and new naming would be misleading as well.

otahirs:
Keep in mind that `Card rent requested` is still saved in the db as `cardLent` and while using this flag, quickevent consider during readout that the card is actually rented, not only that the rent was requested. 
If the runner comes with his own card before the race, i would be much more likely to remove flag thats named `card rented` than `rent requested`.. and with that I will avoid "no thats my card" conversation during readout.
I agree that we can come up with better name, but I stand by the fact that it should not be named `request` when in fact the consequences are broader.

fvacek:
what about this?
[...]

PR #816, opened 2022-10-10
refactor: improve card flags 

otahirs:
#691
image


quickevent/app/quickevent/plugins/Runs/src/cardflagsdialog.h
```diff
@@ -25,6 +25,7 @@ class CardFlagsDialog : public QDialog
 	Ui::CardFlagsDialog *ui;
 	RunsTableModel *m_model;
 	int m_row;
+	void updateStatus();
```
fvacek, line 28: pls move the function declaration from fields part

PR #839, opened 2022-12-19
fix: html characters escape

### LINKED ISSUE #751: Export into xml - need to escape &<>'" characters
When export into xml format (at least startlist) it is needed to escape &<>'" characters.

https://www.xmltutorial.info/xml/special-characters-in-xml/

```
<Organisation>
	<ShortName>RCP</ShortName>
	<Name>RC&P TEAM Bratislava</Name>
</Organisation>
```

Note: to check this, xml is not valid with "&" inside (try to open in ay xml viewer/browser), once it is corrected, xml is valid.

otahirs:
resolve #751

### DIFF
--- a/quickevent/app/quickevent/plugins/Runs/src/runsplugin.cpp
+++ b/quickevent/app/quickevent/plugins/Runs/src/runsplugin.cpp
@@ -1702,7 +1702,7 @@ void RunsPlugin::export_startListClassesHtml()
 						"a",
 						QVariantMap{{"href", "#class_" + tt1_row.value("classes.name").toString()}},
 						tt1_row.value("classes.name"),
-						"&nbsp;"
+						"\0xA0"
 					});
 	}
 	append_list(body, div1);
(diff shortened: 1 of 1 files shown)


otahirs:
`\0xA0` is an UTF-8 character for non-breaking space https://www.compart.com/en/unicode/U+00A0
I have tested the HTML exports and they work.

fvacek:
If it is the same, why we cannot use `&nbsp;`, what is the change?

fvacek:
`"\0xA0"` is IMO no valid HTML entity name, it is even not valid UTF8 code, so it might work, by I afraid, that only by accident. If we want to avoid ampersands in output, because QString::toHtmlString() will interpret them, then we should use something like `"\u00A0"`

otahirs:
`\0xA0`   

image

___
`\u00A0` does render differently  

image


___
I do not plan digging deeper on this issue

fvacek:
It's because "\0xA0" is not valid UTF8 code

fvacek:
why, not use `"\u00A0"`?

fvacek:
resolved by [72415ca](https://github.com/Quick-Event/quickbox/commit/72415ca2bdac9358bf74de56765c3b974d27068e)

PR #873, opened 2023-04-01
fix(OResults): serialize xml uploads, suport for multistage events

otahirs:
1. make sure results and startlist does not upload to the server in the same timeframe. ~~not the best kind of serialization, but the delay should be enough.~~
2. support multi-stage events

~~2. save the `stage_id` when service was started. so change of stage does not affect the periodic exports~~

### DIFF
--- a/quickevent/app/quickevent/plugins/Event/src/services/oresultsclient.cpp
+++ b/quickevent/app/quickevent/plugins/Event/src/services/oresultsclient.cpp
@@ -53,9 +53,7 @@ QString OResultsClient::serviceName()
 
 void OResultsClient::run() {
 	Super::run();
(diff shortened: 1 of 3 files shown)


fvacek:
What about adding callback lambda `on_success` to `sendFile()`?

otahirs:
there has been some major fixes to the iof xml exports and oresults service. can we maybe schedule a (pre-)release of new quickevent version?

fvacek:
> there has been some major fixes to the iof xml exports and oresults service. can we maybe schedule a (pre-)release of new quickevent version?

agree, I'm busy next week, but I'll try to find couple of minutes.

PR #886, opened 2023-04-18
fix: context, multipart leak

otahirs:
fix incorrectly set context object and `multi_part` memory leak

### DIFF
--- a/quickevent/app/quickevent/plugins/Event/src/services/oresultsclient.cpp
+++ b/quickevent/app/quickevent/plugins/Event/src/services/oresultsclient.cpp
@@ -128,8 +128,9 @@ void OResultsClient::sendFile(QString name, QString request_path, QString file, 
 	QUrl url(API_URL + request_path);
 	QNetworkRequest request(url);
 	QNetworkReply *reply = m_networkManager->post(request, multi_part);
+	multi_part->setParent(reply);
 
-	connect(reply, &QNetworkReply::finished, reply, [reply, name, on_success]()
+	connect(reply, &QNetworkReply::finished, this, [=]()
 	{
 		if(reply->error()) {
 			auto err_msg = "OReuslts.eu [" + name + "]: ";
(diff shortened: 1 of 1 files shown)


quickevent/app/quickevent/plugins/Event/src/services/oresultsclient.cpp
```diff
@@ -128,8 +128,9 @@ void OResultsClient::sendFile(QString name, QString request_path, QString file,
 	QUrl url(API_URL + request_path);
 	QNetworkRequest request(url);
 	QNetworkReply *reply = m_networkManager->post(request, multi_part);
+	multi_part->setParent(reply);
 
-	connect(reply, &QNetworkReply::finished, reply, [reply, name, on_success]()
+	connect(reply, &QNetworkReply::finished, this, [=]()
```
fvacek, line 133: It seems like attempt to solve importer crash. I do not think, that this will help. Original version was not leak, IMO. Passing this as context object doesn't do much, since o-result service object lives as long as application itself.

otahirs: the leak was referring to the `multi_part` object not having parent set and therefore never scheduled to be deleted.  

i switched `reply` to `this` for the context parameter, because i have never seen in other code, that the first parameter would also be used in the 3rd context parameter.

JaroslavBeran:
I expect this bug is related to our events during last weekend at Petrohrad.

We were facing a lot of QE crashes during the races.
It happend each 20th up to 50th card read.
When the runner read out his card the DB record was correctly created, but the app immediately crashed without printing receipts.
Windows stopped the app and we had to start the app again then OResults service was either started automatically again or we did it manually.
Then the runner either could read out his card again so new DB record was created and his receipt was printed or we could print the receipt maually for his already created DB record only.

quickevent/app/quickevent/plugins/Event/src/services/oresultsclient.cpp
fvacek: Now I understand, so `multi_part->setParent(reply);` fixes memory leak, context object change is not related.

Context object is used to discard connection + lambda, when context object is deleted. If `reply` is signal sender, then context object is not needed indeed. But, `on_success` might use `this` in implementation, so assigning `this` to context-object might help, even if it is not clearly visible from code.

PR #896, opened 2023-04-24
fix: incompatibility between libpq and libssl

otahirs:


arnost00:
Checked windows installer - it works (PSQL, ORIS).

paukert:
We should make a new release and hide/delete latest release where PostgreSQL does not work.

arnost00:
We need new release, people are calling because 2.6.18 has PSQL bug

fvacek:
New release will be created after this will be build successfully. Unfortunately, this MR doesn't contain QE version bump, I have to make it in new commit.

PR #908, opened 2023-05-02
Update the Ukrainian translation manually

sakhnik:
Hello, I've updated the Ukrainian translation manually. Is this reasonable as I see the translations are usually updated with some weblate?

It's worth noting that QuickEvent is fine-tuned for Latin-based scripts, and hardly usable with Cyrillic-based languages unless I patch up the string collation routine: https://github.com/Quick-Event/quickbox/commit/6177c5e4f8bb3f1cfd6c94cff36fdab78c70979d . Could a more generic collation be implemented as an optional feature?

### DIFF
--- a/libqf/libqfcore/libqfcore-uk_UA.ts
+++ b/libqf/libqfcore/libqfcore-uk_UA.ts
@@ -7,7 +7,7 @@
         <location filename="src/sql/catalog.cpp" line="34"/>
         <location filename="src/sql/catalog.cpp" line="62"/>
         <source>Error getting the sequence nextval(&apos;%1&apos;)</source>
-        <translation>Помилка при отриманні послідовності nextval (&apos;%1&apos;)</translation>
+        <translation>Помилка при отриманні nextval послідовності («%1»)</translation>
     </message>
 </context>
 <context>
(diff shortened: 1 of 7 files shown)


paukert:
> Hello, I've updated the Ukrainian translation manually. Is this reasonable as I see the translations are usually updated with some weblate?

There is no issue with updating translations using pull request :) Weblate is just another way which is probably friendlier for users without experience of PR flow / for users without specialized software.

fvacek:
> It's worth noting that QuickEvent is fine-tuned for Latin-based scripts, and hardly usable with Cyrillic-based languages unless I patch up the string collation routine: [6177c5e](https://github.com/Quick-Event/quickbox/commit/6177c5e4f8bb3f1cfd6c94cff36fdab78c70979d) . Could a more generic collation be implemented as an optional feature?

We are planning to move to Qt6. Collation is made slightly different way here, so some changes will be necessary. I suggest to create separate issue for Cyrillic-based languages. We will need some hints how to implement this correctly.

sakhnik:
> > It's worth noting that QuickEvent is fine-tuned for Latin-based scripts, and hardly usable with Cyrillic-based languages unless I patch up the string collation routine: [6177c5e](https://github.com/Quick-Event/quickbox/commit/6177c5e4f8bb3f1cfd6c94cff36fdab78c70979d) . Could a more generic collation be implemented as an optional feature?
> 
> We are planning to move to Qt6. Collation is made slightly different way here, so some changes will be necessary. I suggest to create separate issue for Cyrillic-based languages. We will need some hints how to implement this correctly.
> 
> I'll review [6177c5e](https://github.com/Quick-Event/quickbox/commit/6177c5e4f8bb3f1cfd6c94cff36fdab78c70979d) meanwhile and it will be merged if it will not affect correct behavior of latin collation.

I'm sorry, that change isn't intended to be merged yet. I wouldn't want to interfere with the Czech language support. I'd just like to attract attention to the issue. Let me open a ticket and take another look at how to fix the collation in a more generic way.

PR #919, opened 2023-05-24
fix: ??? in RunStatus

otahirs:
fix toString methods producing "???" as a result status. Why would we use `???` instead of a default value anyway?  

there is an unsolved question how the runners got `isDisqualified` flag, without any other more concrete flag set in #917

### DIFF
--- a/libquickevent/libquickeventcore/src/runstatus.cpp
+++ b/libquickevent/libquickeventcore/src/runstatus.cpp
@@ -43,8 +43,6 @@ bool RunStatus::isOk() const
 
 int RunStatus::toTime() const
 {
-	if (m_notCompeting)
-		return quickevent::core::og::TimeMs::NOT_COMPETITING_TIME_MSEC;
 	if (m_disqualified) {
 		if (m_missingPunch)
 			return quickevent::core::og::TimeMs::MISPUNCH_TIME_MSEC;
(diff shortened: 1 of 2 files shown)


libquickevent/libquickeventcore/src/runstatus.cpp
```diff
@@ -43,113 +43,103 @@ bool RunStatus::isOk() const
 
 int RunStatus::toTime() const
 {
+	if (m_notCompeting)
```
paukert, line 46: I am not sure if this is correct solution. If the runner is marked as NC and e.g. it is also set m_missingPunch flag, then he should have MissingPunch/MP/... status in results or NC?

arnost00:
I still think that the original order was actually good.
The only thing I didn't expect was that `isDisqualified` would be set separately. `???` was there in case this possibility somehow arises (something like "it shouldn't get here"). 
I agree with the amendment that NC has absolute priority.
And with canceling `???` when the runner only has `isDisqualified` too.

The order of evaluation in my opinion:
  1. Not Competing
  2. Disqualified by org.
  3. MissingPunch
  4. DidNotStart
  5. DidNotFinish
  6. Overtime

jmacura:
Related discussion #669

otahirs:
> ??? was there in case this possibility somehow arises (something like "it shouldn't get here").

I still think it shouldn't get here, but introducing "`null`" value where it is not necessary should be avoided.

Needless to say we should investigate how it got to this state :)

arnost00:
> Needless to say we should investigate how it got to this state :)

Yes, we need to find out how and why.
I started writing the findings into the issue #917

libquickevent/libquickeventcore/src/runstatus.cpp
```diff
@@ -43,113 +43,103 @@ bool RunStatus::isOk() const
...
 
 QString RunStatus::toXmlExportString() const
 {
+	if (m_notCompeting)
+		return QStringLiteral("NotCompeting");
 	if (m_disqualified) {
 		if (m_disqualifiedByOrganizer)
-			return QStringLiteral("Disqualified");
-		else if (m_missingPunch)
+			QStringLiteral("Disqualified");
```
paukert, line 70: Missing `return` statement.

arnost00:
@otahirs look at https://github.com/Quick-Event/quickbox/issues/917#issuecomment-1565681776 
I hope I found how this error occurs.

Issue #373, opened 2019-04-10
Print start list and results on cash printer


fvacek:

fvacek:
Print startlist/results options dialog has new option `page-width`. This can make it work on the cash printer. check version 2.2.4

PR #937, opened 2023-07-26
Fix page layouts & some ReportOptionsDialog improvements

arnost00:
- Fix page layouts for use with receipt printer
- ReportOptionsDialog  
  - add Reset button
  - add select start number for print startlist by classes and for starters

Affects #805, #253, #213, #373

### DIFF
--- a/libquickevent/libquickeventgui/src/reportoptionsdialog.cpp
+++ b/libquickevent/libquickeventgui/src/reportoptionsdialog.cpp
@@ -196,7 +196,6 @@ void ReportOptionsDialog::setOptions(const ReportOptionsDialog::Options &options
 	ui->edPageHeight->setValue(options.pageHeight());
 	ui->edHorizontalMargin->setValue(options.horizontalMargin());
 	ui->edVerticalMargin->setValue(options.verticalMargin());
-	//ui->chkShirinkPageWidthToColumnCount->setChecked(options.isShirinkPageWidthToColumnCount());
 	ui->grpClassFilter->setChecked(options.isUseClassFilter());
 	ui->chkClassFilterDoesntMatch->setChecked(options.isInvertClassFilter());
 	ui->edFilter->setText(options.classFilter());
@@ -233,7 +232,6 @@ ReportOptionsDialog::Options ReportOptionsDialog::options() const
 	opts.setPageHeight(ui->edPageHeight->value());
 	opts.setHorizontalMargin(ui->edHorizontalMargin->value());
 	opts.setVerticalMargin(ui->edVerticalMargin->value());
-	//opts.setShirinkPageWidthToColumnCount(ui->chkShirinkPageWidthToColumnCount->isChecked());
 	opts.setUseClassFilter(ui->grpClassFilter->isChecked());
 	opts.setInvertClassFilter(ui->chkClassFilterDoesntMatch->isChecked());
 	opts.setClassFilter(ui->edFilter->text());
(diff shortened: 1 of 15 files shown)


libquickevent/libquickeventgui/src/reportoptionsdialog.cpp
```diff
@@ -36,14 +39,32 @@ ReportOptionsDialog::ReportOptionsDialog(QWidget *parent)
...
+		ui->grpClassStartSelection->setChecked(false);
+		ui->cbxStartNumber->clear();
+	}
 	connect(ui->btSaveAsDefault, &QPushButton::clicked, [this]() {
 		savePersistentSettings();
 	});
 	connect(this, &ReportOptionsDialog::persistentSettingsIdChanged, [this]() {
 		loadPersistentSettings();
 	});
+    connect(ui->btReset, &QPushButton::clicked, this, &ReportOptionsDialog::resetPersistentSettings);
```
fvacek, line 67: indent by TAB

libquickevent/libquickeventgui/src/reportoptionsdialog.cpp
```diff
@@ -123,13 +146,33 @@ bool ReportOptionsDialog::isStartListPrintStartNumbers() const
...
-QString ReportOptionsDialog::sqlWhereExpression() const
+QString ReportOptionsDialog::sqlWhereExpression(const int stage_id) const
 {
 	const Options opts = options();
-	return sqlWhereExpression(opts);
+	return sqlWhereExpression(opts,stage_id);
 }
 
-QString ReportOptionsDialog::sqlWhereExpression(const ReportOptionsDialog::Options &opts)
+QString ReportOptionsDialog::getClassesForStartNumber(const int number, const int stage_id)
```
fvacek, line 155: I think, that we should rather add `startNUmber` or `startName` or `startId` column to classes.

fvacek (changes requested):

libquickevent/libquickeventgui/src/reportoptionsdialog.cpp
arnost00: Sorry :(

libquickevent/libquickeventgui/src/reportoptionsdialog.cpp
arnost00: Start number is used only in 2 places. 
- Startlist print (for classes & starters)
- XML Startlist export 

I didn't find it necessary to add a column to the db for this.

fvacek: If we plan to start to support more starts, then we should also give user some UI to work with. Use start-code number for this is cryptic and hard to find for user. So I suggest to either support more starts in correct way or nothing.

What user will certainly want is:
1. possibility to define more starts, `courses` table is best place IMO.
2. Possibility to give start some name.

What we need to have this:
1. add one column to `courses` table
2. define one column in UI table

This is IMO more straightforward solution than mangle start-name to code of start station in codes. Also users will be happy, because they will got table column, where they can put start-name/number, whatever they want, on the place where they will be looking for such kind of information.

arnost00: Its for discusion. Maybe from people who wrote to affected issues.
For me :
 - Start codes/names are imported from courses. I've never didn't need to change it. I only use it from courses. 
 - And I only need it to print startlist for starters.
 - I don't need to edit or manage it in any way.

arnost00: I wrote this because I need it for WMOC.

fvacek: Ok, let to it this way for now. I realized that, if we introduce new DB column, then we can have different start-number defined in `courses` table and different start-code in `classcodes`. This is something also not wanted.

PR #955, opened 2023-12-25
Fix issue #954, add missing ReportOptionsDialog PageLayout call

arnost00:


quickevent/app/quickevent/plugins/Runs/src/runswidget.cpp
```diff
@@ -220,8 +220,21 @@ void RunsWidget::settleDownInPartWidget(quickevent::gui::PartWidget *part_widget
 		auto *a = new qfw::Action(tr("&Competitors with rented cards"));
 		connect(a, &qfw::Action::triggered, [this]() {
 			qff::MainWindow *fwk = qff::MainWindow::frameWork();
+			quickevent::gui::ReportOptionsDialog dlg(fwk);
```
fvacek, line 223: please, bump QE version

fvacek (changes requested):

quickevent/app/quickevent/plugins/Runs/src/runswidget.cpp
arnost00: No problem.
For this PR and PR #956 together or for both independently?

fvacek: for both pls

arnost00: Done

PR #956, opened 2023-12-25
Fix issue #953, better computation of start time for Emma Client (racom type)

arnost00:
Fix when runs.startTimeMs is null


fvacek:
Prosim povysit verzi QE

arnost00:
> Prosim povysit verzi QE

Povyseno

PR #991, opened 2025-01-17
feat: nationality support

otahirs:
this adds basic support for Nationality (country code)   
the motivation is to have national flags shown on [oresults.eu](https://oresults.eu), as quickevent lacks those in the exports in comparison to meos or kramer

#### normal events
normal events gets nationality info from ORIS as the endpoint `https://oris.orientacnisporty.cz/API/?format=json&method=getEventEntries&eventid=%1` contains Nationality data, even thought they need to be converted into IOC format https://github.com/international-orienteering-federation/datastandard-v3/blob/24eb108e4c6b5e2904e5f8f0e49142e45e2c5230/IOF.xsd#L4224
#### relay
relay endpoint for syncing entries does not have nationality data, so no automatic support there yet for importing it

### Export / Import
implemented for normal/relay iof xml import and exports  
xmlimporter also touched


quickevent/app/quickevent/plugins/Oris/src/alphatoioc.h
```diff
@@ -0,0 +1,250 @@
+#ifndef ALPHATOIOC_H
+#define ALPHATOIOC_H
+
+#include <QMap>
+#include <QString>
+
+QMap<QString, QString> isoAlpha2ToIoc = {
```
arnost00, line 7: Ja nevim, fixni seznam kodů v zdrojaku. Jak budeme řešit když dojde k nějaké opravě nebo úpravě kodů?
Názvy zemí se v průběhu času vyvíjejí.

arnost00:
@otahirs, do I need to go through the code for the races from Eventor?

arnost00:
@otahirs, how did you write it to the database? I don't see any change in the db definition (DbSchema.qml). But from the code it seems to me that you added some columns.

quickevent/app/quickevent/plugins/Oris/src/alphatoioc.h
otahirs: I suppose we dont use any packaging system in this c++ project that would handle this. Git submodule for such a small thing would be overkill. The countries participating in orienteering are hardly going to change to have a meaningful impact, but I am open to suggesting how to handle this better.

otahirs:
should I be targeting `qe3` instead of `master`?

arnost00:
@otahirs, you need to increase the DB version in : `quickevent\app\quickevent\src\application.cpp` method `Application::dbVersion()`

arnost00:
> should I be targeting `qe3` instead of `master`?

I'm still creating new features for both `master` and `qe3`

fvacek:
> should I be targeting `qe3` instead of `master`?

yes please

fvacek (approved):

arnost00:
But also, the import of registration from ORIS does not contain data on nationality, so manual entry to the event will also be without nationality, so we will achieve another data inconsistency.

But without expanding the api export format (registration) from ORIS, this probably can't be solved any better.

PR #997, opened 2025-04-28
QE3 - added OFeed platform support

lukaskett:
The service features:
- manually upload data - start list, results
- regular results data upload (according to the set interval)
- send changes - edited competitor, new competitor, read out data

TODO:
- [ ] Process changes that come from start/OChecklist app (origin=START) - changed status, card number or note
- [ ] Add a more sophisticated way to distinguish between new and edited competitors (compares competitor id and actual sequence value at this moment) because both actions invokes the same db event.

### DIFF
--- a/quickevent/app/quickevent/plugins/Event/src/services/ofeed/ofeedclient.cpp
+++ b/quickevent/app/quickevent/plugins/Event/src/services/ofeed/ofeedclient.cpp
@@ -1,0 +1,367 @@
+#include "ofeedclient.h"
+#include "ofeedclientwidget.h"
+
+#include "../../eventplugin.h"
+
+#include <qf/qmlwidgets/framework/mainwindow.h>
+#include <qf/qmlwidgets/dialogs/dialog.h>
+
+#include <qf/core/log.h>
+#include <plugins/Runs/src/runsplugin.h>
+#include <plugins/Relays/src/relaysplugin.h>
+#include <quickevent/core/si/checkedcard.h>
+#include <qf/core/utils/htmlutils.h>
+#include <qf/core/sql/query.h>
+
+#include <QCoreApplication>
+#include <QDir>
+#include <QFile>
+#include <QHttpPart>
+#include <QNetworkAccessManager>
+#include <QNetworkReply>
+#include <QSettings>
+#include <QStandardPaths>
+#include <QTextStream>
+#include <QTimer>
+#include <regex>
+
+namespace qfc = qf::core;
+namespace qfw = qf::qmlwidgets;
+namespace qfd = qf::qmlwidgets::dialogs;
+namespace qfs = qf::core::sql;
+using qf::qmlwidgets::framework::getPlugin;
+using Event::EventPlugin;
+using Relays::RelaysPlugin;
+using Runs::RunsPlugin;
+
+namespace Event {
+namespace services {
+
+OFeedClient::OFeedClient(QObject *parent)
+    : Super(OFeedClient::serviceName(), parent)
+{
+	m_networkManager = new QNetworkAccessManager(this);
+	m_exportTimer = new QTimer(this);
+    connect(m_exportTimer, &QTimer::timeout, this, &OFeedClient::onExportTimerTimeOut);
+    connect(this, &OFeedClient::settingsChanged, this, &OFeedClient::init, Qt::QueuedConnection);
+    connect(getPlugin<EventPlugin>(), &Event::EventPlugin::dbEventNotify, this, &OFeedClient::onDbEventNotify, Qt::QueuedConnection);
+}
+
+QString OFeedClient::serviceName()
+{
+    return QStringLiteral("OFeed");
+}
+
+void OFeedClient::run() {
+	Super::run();
+	exportStartListIofXml3([this](){exportResultsIofXml3();});
+	m_exportTimer->start();
+}
+
+void OFeedClient::stop() {
+	Super::stop();
+	m_exportTimer->stop();
+}
+
+void OFeedClient::exportResultsIofXml3()
+{
+	int current_stage = getPlugin<EventPlugin>()->currentStageId();
+	bool is_relays = getPlugin<EventPlugin>()->eventConfig()->isRelays();
+
+	QString str = is_relays
+			? getPlugin<RelaysPlugin>()->resultsIofXml30()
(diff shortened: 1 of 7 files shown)


quickevent/app/quickevent/plugins/Event/src/services/ofeed/ofeedclient.cpp
```diff
@@ -0,0 +1,661 @@
...
+
+            // Add eventId field
+            QHttpPart event_id_part;
+            event_id_part.setHeader(QNetworkRequest::ContentDispositionHeader, QVariant("form-data; name=\"eventId\""));
+            event_id_part.setBody(eventId().toUtf8());
+            multi_part->append(event_id_part);
+
+            // Add xml content with fake filename that must be present
+            QHttpPart file_part;
+            file_part.setHeader(QNetworkRequest::ContentDispositionHeader, QVariant("form-data; name=\"file\"; filename=\"uploaded_file.xml\""));
```
fvacek, line 149: Not sure, if file-name should be sent as form-data, seem to me to be copy-pasted from o-results. I suggest to encode file-name to url query, like `POST https://ofeed-host/rest/v1/upload/iof?filename=uploaded_file.xml`

quickevent/app/quickevent/plugins/Event/src/services/ofeed/ofeedclient.cpp
```diff
@@ -0,0 +1,661 @@
...
+            exportResultsIofXml3();
+        }
+
+        void OFeedClient::loadSettings()
+        {
+            Super::loadSettings();
+            init();
+        }
+
+        void OFeedClient::sendFile(QString name, QString request_path, QString file, std::function<void()> on_success)
```
fvacek, line 130: parameter name is never used

quickevent/app/quickevent/plugins/Event/src/services/ofeed/ofeedclient.cpp
```diff
@@ -0,0 +1,661 @@
...
+                return;
+            }
+            bool update_competitor = true;
+            // Handle a new competitor - use the value of the sequence and check current value?
+            // TODO: fix schema name
+            // qf::core::sql::Connection::forName();
+            // qf::core::utils::ConfigCLIOptions
+            // cliOptions()->eventName()
+            qf::core::sql::Query q_schema;
+            q_schema.exec("select current_schema", qf::core::Exception::Throw);
```
fvacek, line 403: This is too cumbersome way to find out if competitor was edited or inserted, we should better add this flag to DB_EVENT payload data

fvacek: This also will not work with SQLite

quickevent/app/quickevent/plugins/Event/src/services/ofeed/ofeedclient.cpp
```diff
@@ -0,0 +1,661 @@
...
+
+                // Time
+                int running_time = INT_INITIAL_VALUE;
+                QVariant running_time_variant = q.value(QStringLiteral("timeMs"));
+                if (!running_time_variant.isNull())
+                {
+                    running_time = running_time_variant.toInt();
+                }
+
+                // Status
```
fvacek, line 499: There is a `RunStatus` utility class in libquickevent for this already

fvacek (changes requested):
Please fix indentation and read also QE coding style https://github.com/Quick-Box/quickevent/wiki/Code-style

quickevent/app/quickevent/plugins/Event/src/services/ofeed/ofeedclient.cpp
lukaskett: is used in log

quickevent/app/quickevent/plugins/Event/src/services/ofeed/ofeedclient.cpp
lukaskett: `RunStatus` is not operating with `Inactive` status which is important as initial state for the platform

quickevent/app/quickevent/plugins/Event/src/services/ofeed/ofeedclient.cpp
lukaskett: agree, a new db event will solve it in much better way, but adding a new db event was too big task for me to be honest.

quickevent/app/quickevent/plugins/Event/src/services/ofeed/ofeedclient.cpp
fvacek: What `Inactive` exactly means? 
```
innactive ==  RunStatus WHERE all fields are false
```
in my understanding

lukaskett: From IOF.xsd, [link](https://github.com/international-orienteering-federation/datastandard-v3/blob/24eb108e4c6b5e2904e5f8f0e49142e45e2c5230/IOF.xsd#L2952)
```xml
<xsd:enumeration value="Inactive">
        <xsd:annotation>
          <xsd:documentation>
            Has not yet started.
          </xsd:documentation>
        </xsd:annotation>
</xsd:enumeration>
```
so it makes sense that all fields are false

quickevent/app/quickevent/plugins/Event/src/services/ofeed/ofeedclient.cpp
lukaskett: I made additional tests (with and without filename) and it looks that filename must be send as form-data. The filename attribute in the Content-Disposition header is typically required for the server to interpret the data as a file upload.

quickevent/app/quickevent/plugins/Event/src/services/ofeed/ofeedclient.cpp
fvacek: I'll add one

quickevent/app/quickevent/plugins/Event/src/services/ofeed/ofeedclient.cpp
fvacek: Then `RunStatus` can be used, we can introduce `RunStatus::isInactive()`

fvacek:
moved to https://github.com/Quick-Box/quickevent/pull/1009

PR #1011, opened 2025-05-20
Fix Clang tidy errors

fvacek:


quickevent/app/quickevent/plugins/Event/src/eventplugin.cpp.orig
arnost00, line 1: Accidentally added .orig file.

quickevent/app/quickevent/plugins/Classes/src/drawing/ganttscene.cpp
arnost00, line 1: : Super(parent), 
There remains a comma at the end of line.
Build fail on it.

PR #1027, opened 2025-09-06
Dark theme support

fvacek:


fvacek:
image

arnost00:
Entire window in dark mode – some parts are hard to read (highlighted) : 
qe_3_darkmode2

fvacek:
> There are still parts that are not very readable in dark mode.
> 
> There are changes to the appearance of light mode. Was that intentional?

No, it is definitely not an intention, I wonder, how this might happen, did you run `cmake --install` before testing?

PR #1028, opened 2025-09-07
Rename qmlwidgets namespace to gui

fvacek:


arnost00 (approved):
Looks good. Only need to solve conflicts in tablemodel before merge.

Issue #979, opened 2024-06-18
Start list for starters with more starts


xkenia:
When print start list for starterts with more starts and check Class filter (for classes) it do not use Class filter - start number selection.
When try to print starters startlist for all classes for Start 2 (check Class filter - start number selection and select Start 2) and have checked Class filter (with empty filter field) it will print all classes for all starts, not only for selected Start 2.

fvacek:
please , attach the `qbe` file.

lukaskonig:
[sobota_cil_v1.zip](https://github.com/user-attachments/files/22208320/sobota_cil_v1.zip)

fvacek:
fixed in `QE 3.4.6`

PR #1036, opened 2025-10-01
Fix Start Number selection in reports #979

fvacek:
When print start list for starterts with more starts and check Class filter (for classes) it do not use Class filter - start number selection.
When try to print starters startlist for all classes for Start 2 (check Class filter - start number selection and select Start 2) and have checked Class filter (with empty filter field) it will print all classes for all starts, not only for selected Start 2.


arnost00 (approved):
Nice work. Thanks.

Issue #1038, opened 2025-10-02
Some courses invisible in course edit form


LadaHo:
QE 3.4.6, win11
Some courses (D12D, D14D, D18C, D20C, ...) imported from IOF XML 3.0 are not visible in course edit form.
Classes are imported all. All classes have the correct courses assigned.

LadaHo:
Hi,

XML with data from PurplePen attached

Have a nice day

Ladislav Hovorka
------------------------
+420 724 524 867
***@***.***


čt 2. 10. 2025 v 15:25 odesílatel Richard Patek ***@***.***>
napsal:

lukaskett:
Hi, I made a short test with Purple Pen courses on mentioned version `3.4.6` and everything worked as expected. Could you please provide screenshots, source xml file (I can't see provided one) or exported qbe?

fvacek:
I guess that invisible courses do not have runners in corresponding class.

arnost00:
I tested it with ORIS race 9098. And it look like as @fvacek said.

<img width="798" height="378" alt="Image" src="https://github.com/user-attachments/assets/a0ebd690-da92-4ff6-ab62-45ba9f838c87" />
<img width="845" height="416" alt="Image" src="https://github.com/user-attachments/assets/187a23c2-351e-4d50-a5ae-9ddb81397fee" />
<img width="396" height="263" alt="Image" src="https://github.com/user-attachments/assets/951ce955-e41c-44ef-b9e9-cd13e86e7aae" />

PR #1040, opened 2025-10-02
Some courses invisible in course edit form

fvacek:
#1038


arnost00 (approved):
Formulář_2025-10-03_11-53-32
Looks good.

PR #1045, opened 2025-10-20
Show relays courses

fvacek:


arnost00:
@fvacek Do you know why the builds in Actions didn't start?

arnost00:
Tested in relay event & single event. On local build.

PR #1055, opened 2025-10-23
Set tatus DNS == false after card read #1049

fvacek:


arnost00 (approved):
Tested. Works. Thanks.

PR #1080, opened 2026-01-05
Add missing support for import relay courses from IOF XML 3.0

arnost00:
Add missing support for import relay courses from IOF XML 3.0, mentionned in #972


arnost00:
@Osin01 please test it with XML files from PurplePen.

PR #1096, opened 2026-02-24
feat: macos support

otahirs:
necessary edits and build pipeline for MacOS support

---
bonus:
## guide for setting up driver for sportident SI unit:

kudos to original guide on <https://sdenier.github.io/Geco/begin/faq.html>

1. install driver for the chip from silicon labs that is inside sporident units 
   <https://www.silabs.com/software-and-tools/usb-to-uart-bridge-vcp-drivers?tab=downloads>
2. allow all permissions during the installation, more guidance here: <https://www.hobbyeagle.com/download/docs/usb-installation-guide-for-mac/>

   especially in this step :arrow_down: don’t just click `ok` or the driver will not get required permissions, click open system settings and flick the driver toggle to allow it
image

3. the driver is installed but we need to change its config to recognize the sporident device

   edit the driver config file in your text editor

```bash
/Applications/CP210xVCPDriver.app/Contents/Library/SystemExtensions/com.silabs.cp210x.dext/Info.plist
```

search for section `CP210x USB to UART Bridge`
and change:

```bash
<key>idProduct</key>
<integer>60000</integer>
```

to:

```bash
<key>idProduct</key>
<integer>32778</integer>
```

`32778` is id pro sportident chip

4. restart MacOS
5. when the sportident unit is connected, you should now find the device `/dev/tty.SLAB_USBtoUART`

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,12 +70,11 @@ sudo usermod -aG lp $USER
 #### macOS
 1. download the DMG file from https://github.com/Quick-Event/quickbox/releases or github action build
 2. open the DMG and drag QuickEvent to the Applications folder
-3. **Important**: macOS will block the app with "QuickEvent is damaged" message because it's not code signed
-4. remove the quarantine attribute:
+3. **Important**: macOS will block the app with "QuickEvent is damaged" message because it's not code signed with apple developer account. To remove the quarantine attribute, run:
 ```sh
 xattr -cr /Applications/quickevent.app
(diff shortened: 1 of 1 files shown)


otahirs:
Hi, you see this error because the app is not signed using apple developer account.

Just run `xattr -d com.apple.quarantine /Applications/quickevent.app` and you should be able to run the app.

Issue #1110, opened 2026-04-08
QE crashes after clicking File -> Import -> ORIS -> Update one-time clubs


lukaskett:
When trying to update one-time clubs for [Velikonoce ve skalách 2026 - E1](https://oris.ceskyorientak.cz/Zavod?id=9665) QE crashes.

fvacek:
How this can be reproduced? Do I need event key for this?

lukaskett:
> How this can be reproduced? Do I need event key for this?

Yes, you need an eventkey for event that has over 99 one-time clubs.

I am in touch with Bestik, he will probably add a new request parameter for club abbreviation. The issue has two parts:
- 1. QE crashes because of unexpected response format -> should be fixed
- 2. ORIS it not able to distinguish between e.g.  id 144 of CSOS and one-time club at this moment, because the same id parameter is shared for id and club abbreviation.

fvacek:
Unfortunately, I'm not able to reproduce app crash with current `main` branch on linux. Can anybody try to reproduce this on windows? The event-key will be sent by email on demand.

lukaskett:
> Unfortunately, I'm not able to reproduce app crash with current `main` branch on linux. Can anybody try to reproduce this on windows? The event-key will be sent by email on demand.

Were you importing data from Oris for event 9665, inserted eventkey and then tried to update one-time clubs: `File -> Import -> ORIS -> Update one-time clubs`?

It was officialy reported for **3.4.23** and i tried for the `main` branch as well (on windows).

fvacek:
Then it is bug reproducible on windows only, can you run it in debugger and send me backtrace?

lukaskett:
> Then it is bug reproducible on windows only, can you run it in debugger and send me backtrace?

Sure, how can i do that? It's not mentioned in the [Developers guide](https://github.com/Quick-Box/quickevent/wiki/Developers-guide).

fvacek:
The easiest way is to push debug button in QTCreator. What IDE are you using for QE development?

lukaskett:
> The easiest way is to push debug button in QTCreator. What IDE are you using for QE development?

ok, I'll try QTCreator.

I realized one more thing that there are also clubs that are neither one-time members nor members of ČSOS and are permanently registered with Oris (foreign clubs), e.g. [26D ](https://oris.ceskyorientak.cz/API/?format=json&method=getClub&id=26D)OLV Potsdam. That could the source of the error as well.

[...]

PR #1117, opened 2026-04-17
1110 update one time clubs

fvacek:
#1110


quickevent/app/quickevent/plugins/Oris/src/orisimporter.cpp
```diff
@@ -33,7 +31,7 @@
 #include <QUrl>
 #include <QInputDialog>
 #include <QMessageBox>
-#include <set>
```
arnost00, line 36: My build on Windows failed if ``#include <set>`` not there.

lukaskett:
I tried [build](https://github.com/Quick-Box/quickevent/actions/runs/24577210394) and unfortunately it's still not working - the app crashes.

fvacek:
> I tried [build](https://github.com/Quick-Box/quickevent/actions/runs/24577210394) and unfortunately it's still not working - the app crashes.

I did it as well, without crash. Can @arnost00 test this too, pls?

arnost00:
> I did it as well, without crash. Can @arnost00 test this too, pls?

I can build it & test it. But I don't have ORIS key to problematic event.

fvacek:
> > I did it as well, without crash. Can @arnost00 test this too, pls?
> 
> I can build it & test it. But I don't have ORIS key to problematic event.

I've sent it to you by email.

arnost00:
Problem with DB creation - QStringList DbSchema::loadCreateDbSqlScript return one line 
Problem is line :
auto lines = sql.split(";\n");
Because in script file ends are : ";\r\n"

arnost00:
For Windows is quick fix ``auto lines = sql.split(";\r\n");`` in function ``DbSchema::loadCreateDbSqlScript``
But I'm not sure if it is fix for unix/mac ?

arnost00:
> I did it as well, without crash. Can @arnost00 test this too, pls?

I've tested it (with DB creation quick fix).
It works for me. But never update all one-timed clubs in this race.
obrazek

But this looks like problem with ORIS subscribed in #1110 by @lukaskett

arnost00:
In this version of code progress info doesn't make sence :
```
		int items_processed = 0;
		int items_count = missing_clubs.size();
		fwk->showProgress(tr("Importing one-time clubs"), 0, items_count);
		qfLogScope("importClubs");
		for (auto &abbr : missing_clubs) {
			getAndImportClub(abbr, event_key);
			items_processed++;
		}

		fwk->hideProgress();
		qfInfo() << "Import of"<< items_processed << "new one-time clubs started...";
```
I don't see any progress, no information about how many clubs was found a inserted into DB.


In previous version has every new club log line :
```
2026-04-21T17:05:06.339	Info		Get: https://oris.ceskyorientak.cz/API/?format=json&method=getClub&id=037&eventkey=XXXX OK: true	orisimporter.cpp	127		
```

arnost00:
I've tested with old code - progress shown. 
With new code is nothing to see. I don't know result of this action.

fvacek:
> Problem with DB creation - QStringList DbSchema::loadCreateDbSqlScript return one line Problem is line : auto lines = sql.split(";\n"); Because in script file ends are : ";\r\n"

But `create_db_sqlite.sql` is created with ';\n'. I've no idea when line ends are changed to ';\r\n' on windows. :(

My suspicion is that you have set `core.autocrlf` to `true` in your git.

arnost00:
I try to install 2 buids from QE Actions. Both after db_scripts update, and result are the same.

Quickevent without parameters only generate create_db_psql.sql & create_db_sqlite.sql in install directory.
If I try to add log ... quickevent.exe -v 2> C:\QuickEvent\quickevent.log, result is in attached ...
[quickevent.log](https://github.com/user-attachments/files/26947844/quickevent.log)

Same as @lukaskett wrote to issue #1109

arnost00:
> My suspicion is that you have set `core.autocrlf` to `true` in your git.

Yes, I have.

But better solution is :``f.open(QFile::ReadOnly | QIODevice::Text)`` in dbschema.cpp

Qt docs say .. Qt will translate the end-of-line characters if you pass QIODevice::Text to mode

arnost00:
Db creation & QE Start in Release fixed in #1124

fvacek:
I'm afraid, that event token has expired 20.4.

lukaskett:
> I'm afraid, that event token has expired 20.4.

The event key is not expiring, I guess, only data are deleted and the response is empty. But the result is unfortunately the same. We will have to wait for another event with more than 100 one-time clubs. World Cup in Vyšší Brod has big potential.

PR #1120, opened 2026-04-19
Improved relays import from ORIS - use API

arnost00:
Reworked from CSOS TXT to ORIS API. 
TXT has fixed column length, and use cp-1250.
Preparation for fix issue #884


fvacek:
https://github.com/Quick-Box/quickevent/pull/1117 should be merged first

arnost00:
Ok. I'll wait

fvacek (approved):

PR #1125, opened 2026-04-22
Fix create db for psql

fvacek:


arnost00 (approved):
Tested. Worked with PSQL17. Thanks

arnost00:
Maybe version bump to 3.5.4 ?

Issue #1127, opened 2026-04-24
Invalid iof xml results


lukaskett:
I noticed that exported xml results are not valid, `Date` element is empty and iof xml validation fails.
```xml
<StartTime>
    <Date></Date>
    <Time>10:30:00.000</Time>
</StartTime>
```
[pbm_3jml_2026_12_export_v350.results-iof3.xml](https://github.com/user-attachments/files/27068161/pbm_3jml_2026_12_export_v350.results-iof3.xml)

Effected builds:
- https://github.com/Quick-Box/quickevent/actions/runs/24803825379
- https://github.com/Quick-Box/quickevent/actions/runs/24391712588

arnost00:
Issue talks about element ``ResultList\Event\StartTime``

Element ``ResultList\ClassResult\PersonResult\Result\StartTime`` has full date/time 
```
				<StartTime>2026-04-18T10:58:00+02:00</StartTime>
				<FinishTime>2026-04-18T11:15:58+02:00</FinishTime>
```

Just to clarify.

arnost00:
It was there in QE version 3.4.8 ... ``<Date>2025-10-18</Date>``

arnost00:
@lukaskett check it.

lukaskett:
> [@lukaskett](https://github.com/lukaskett) check it.

works as expected, thanks

lukaskett:
when using `3.5.7` for relays, i face the same issue with empty `Date` element:
```xml
<Event>
		<Id type="ORIS">10128</Id>
		<Name>Žaket Pražský pohár žactva - Klacky open</Name>
		<StartTime>
			<Date></Date>
			<Time>16:45:00.000</Time>
		</StartTime>
		<Official type="director">
			<Person>
				<Name>
					<Family>Rusý</Family>
					<Given>Tomáš</Given>
				</Name>
			</Person>
		</Official>
		<Official type="mainReferee">
			<Person>
				<Name>
					<Family>Kettner</Family>
					<Given>Lukáš</Given>
				</Name>
			</Person>
		</Official>
	</Event>
```

arnost00:
I think, it can be closed.
PR #1173 was merged.

lukaskett:
merged

PR #1128, opened 2026-04-25
Fix event date/time  in XML3.0 result export

arnost00:
Fix event date/time in XML3.0 result export
Use same logic as in startlist
Fix issue #1127


lukaskett (approved):
fine thanks, good to apply the same approach as for the statrtlist

fvacek (approved):

PR #1135, opened 2026-05-01
Fix direct receipt printing layout on Windows (overlapping text / blanks)

ljezek:
## Bug
- `ReceiptsPrinter::printReceipt()` passed the `QPrinter` as paint device to
  `ReportProcessor` for report layout. On Windows, `QPrinter::logicalDpiX()`
  returns a different value before vs. after `QPainter::begin()` is called on
  it, causing a DPI mismatch between the layout phase (pre-begin) and the draw
  phase (post-begin). The result was overlapping text lines and large blank gaps
  in directly-printed receipts.
    * Work-around was using no scaling (100%) - #1012

Left image was generated by `Print Receipt` (bad), right one by `Show Receipt`.

image

## Fix

- Fixed by using the main window (`QWidget`) as the `ReportProcessor` paint
  device, giving layout a stable screen DPI — exactly what `ReportViewWidget`
  does when rendering the receipt preview. `ReportPainter` still draws on the
  `QPrinter`; all stored `renderedRect` coordinates are in mm and are correctly
  converted to printer pixels at draw time.
  * This only affects the bad `Print Receipt` path.

Validation on Windows (Print Receipt to PDF Printer & to Epson TM-T70):
image

## Credits 
Credits go to @casidlo for pointing me to the GH Issue and work-around; and to
Claude Code for the fix (which I verified locally on Windows 11).

I'm not experiencing the other scaling/blurry issues so can't confirm it fixes the full
issue.

### DIFF
--- a/quickevent/app/quickevent/plugins/Receipts/src/receiptsprinter.cpp
+++ b/quickevent/app/quickevent/plugins/Receipts/src/receiptsprinter.cpp
@@ -50,6 +50,7 @@ bool ReceiptsPrinter::printReceipt(const QString &report_file_name, const QVaria
 	ReceiptsSettings settings;
 	QPrinter *printer = nullptr;
 	QPaintDevice *paint_device = nullptr;
+	qff::MainWindow *fwk = qff::MainWindow::frameWork();
 	if(settings.printerTypeEnum() == ReceiptsSettings::PrinterType::GraphicPrinter) {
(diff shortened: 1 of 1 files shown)


fvacek (approved):

fvacek (commented):
@ljezek  Thank you very much 👍

PR #1138, opened 2026-05-03
Implement new DB changes API

fvacek:
I'm not able to finish https://github.com/Quick-Box/quickevent/pull/1069 in near future, but DB cange API is finished already. 

@lukaskett please check if this API fits also needs of O-feed

Usage example:
```c++
QObject::connect(&app, &Application::dbRecChng, &app, [](const qf::core::sql::RecChng &recchng) {
	auto dump_map = [](const QVariantMap &m) {
		QStringList rows;
		for (const auto &[k, v] : m.asKeyValueRange()) {
			rows << QStringLiteral("%1 -> %2").arg(k).arg(v.toString());
		}
		return rows.join('\n');
	};
	qfInfo() << "REC-CHNG table:" << recchng.table
			 << "op:" << qf::core::sql::RecChng::recopToString(recchng.op)
			 << "id:" << recchng.id
			 << "record:" << dump_map(recchng.record);
});
```

### DIFF
--- a/libqf/libqfcore/CMakeLists.txt
+++ b/libqf/libqfcore/CMakeLists.txt
@@ -13,6 +13,8 @@ add_library(libqfcore SHARED
     src/sql/query.cpp
     src/sql/querybuilder.cpp
     src/sql/transaction.cpp
+	src/sql/recchng.h src/sql/recchng.cpp
+	src/sql/qxsql.h src/sql/qxsql.cpp
 
     src/utils/clioptions.cpp
     src/utils/crypt.cpp
--- a/libqf/libqfcore/CMakeLists.txt
+++ b/libqf/libqfcore/CMakeLists.txt
@@ -13,7 +13,7 @@ add_library(libqfcore SHARED
     src/sql/query.cpp
     src/sql/querybuilder.cpp
     src/sql/transaction.cpp
-	src/sql/recchng.h src/sql/recchng.cpp
+	src/sql/qxrecchng.h src/sql/qxrecchng.cpp
 	src/sql/qxsql.h src/sql/qxsql.cpp
 
     src/utils/clioptions.cpp
--- a/libqf/libqfcore/include/qf/core/sql/qxrecchng.h
+++ b/libqf/libqfcore/include/qf/core/sql/qxrecchng.h
@@ -1,0 +1,1 @@
+#include "../../../../src/sql/qxrecchng.h"
--- a/libqf/libqfcore/include/qf/core/sql/qxsql.h
+++ b/libqf/libqfcore/include/qf/core/sql/qxsql.h
@@ -1,0 +1,1 @@
+#include "../../../../src/sql/qxsql.h"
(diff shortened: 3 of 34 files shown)


lukaskett:
> I'm not able to finish #1069 in near future, but DB cange API is finished already.
> 
> @lukaskett please check if this API fits also needs of O-feed
> 
> Usage example:
> 
> ```c++
> QObject::connect(&app, &Application::dbRecChng, &app, [](const qf::core::sql::RecChng &recchng) {
> 	auto dump_map = [](const QVariantMap &m) {
> 		QStringList rows;
> 		for (const auto &[k, v] : m.asKeyValueRange()) {
> 			rows << QStringLiteral("%1 -> %2").arg(k).arg(v.toString());
> 		}
> 		return rows.join('\n');
> 	};
> 	qfInfo() << "REC-CHNG table:" << recchng.table
> 			 << "op:" << qf::core::sql::RecChng::recopToString(recchng.op)
> 			 << "id:" << recchng.id
> 			 << "record:" << dump_map(recchng.record);
> });
> ```

Thanks for details, I'll check it and let you know.

lukaskett:
@fvacek
Notes:
- table `competitors` is not supporting **RecChng**, `Super::saveData()` in `CompetitorDocument` is still used
-  direct table edits (e.g. **Runs** tab) are not detected, `RunsTableModel::postRow()` is used
- changes from another pc in the network is **not detected** (it's just local QT signal)
 
I'm not sure we're on the same page. I expected it to work like the existing database events, but with the difference that it would handle the actual data changes rather than just the competitor IDs.

As far as I understand, this change isn't currently much more suitable for OFeed than the existing database events.

lukaskett:
Related #1063

fvacek:
>     * table `competitors` is not supporting **RecChng**, `Super::saveData()` in `CompetitorDocument` is still used
@lukaskett 
This is not clear to me, what is not working.
image

fvacek:
@lukaskett 
> changes from another pc in the network is not detected (it's just local QT signal)

Changes are not currently propagated to other QE instances. My understanding is that o-feed is interested on local changes, which are sent to o-feed and then distributed via o-feed service to other clients.

lukaskett:
> @lukaskett This is not clear to me, what is not working. image

sorry, i overlooked, it works.

lukaskett:
> Changes are not currently propagated to other QE instances. My understanding is that o-feed is interested on local changes, which are sent to o-feed and then distributed via o-feed service to other clients.

Yes, but that would require the OFeed service to be running on every QE instance. The idea was to have one pc for sending changes to OFeed (and process changes from other origins), which is more robust than having to manage the service on every QE instance in my point of view.

fvacek:
> > Changes are not currently propagated to other QE instances. My understanding is that o-feed is interested on local changes, which are sent to o-feed and then distributed via o-feed service to other clients.
> 
> Yes, but that would require the OFeed service to be running on every QE instance. The idea was to have one pc for sending changes to OFeed (and process changes from other origins), which is more robust than having to manage the service on every QE instance in my point of view.

Well, I can add DB_EVENT for this, but I rather like to make short meeting with you to make things clear.

fvacek:
@lukaskett events are passed via PSQL notifications

Issue #1113, opened 2026-04-15
Mac version external database error


sergiogpajares:
Hi,

Whenever I try to connect to an external database on the Mac version I get the following error message: "Connect Database Error: Driver not loaded Driver not loaded"

<img width="444" height="157" alt="Image" src="https://github.com/user-attachments/assets/db01bfe6-192e-4c8a-877c-017eb2cbe099" />

It does not matter if the database is on the localhost or in another computer.

fvacek:
It seems that Mac installer to not contain all necessary files. This message probably means that `libqsqlpsql.so` has some missing dependency. Installing some PostgreSql GUI tool might help, but I do not have Mac and do not know much about this system. 

Maybe @lukaskett can help.

lukaskett:
@fvacek I am Windows user, @otahirs is the guy who got QuickEvent working on a Mac 👍

otahirs:
see https://github.com/Quick-Box/quickevent?tab=readme-ov-file#macos
> For PostgreSQL database connections, you may need to install PostgreSQL client libraries:
> `brew install libpq`

lukaskett:
@fvacek misclick?

fvacek:
> [@fvacek](https://github.com/fvacek) misclick?

No, @sergiogpajares  asked me to reopen issue.

otahirs:
if "_it works on my machine_", than its probably a bundling issue, i could try to find it. in the meantime, @sergiogpajares can try to install `qt-postgresql` or the whole  `qt` bundle with `brew` too see if the error originates from missing qt dependency

sergiogpajares:
It does not fix the issue right away.

This might be related to the problem. The error I see from the terminal is quite confusing:
```
qt.sql.qsqldatabase: QSqlDatabase: can not load requested driver 'QPSQL', available drivers: QSQLITE QODBC QPSQL QMIMER
```

If you check the GitHub Actions output for macdeployqt, you’ll see the following errors:
```
ERROR: "error: /Applications/Xcode_16.4.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/otool-classic: can't open file: /Applications/Postgres.app/Contents/Versions/14/lib/libpq.5.dylib (No such file or directory)\n"
ERROR: no file at "/Applications/Postgres.app/Contents/Versions/14/lib/libpq.5.dylib"
[...]

PR #1139, opened 2026-05-04
Fix MacOS postgres driver bundle issue

sergiogpajares:
Dynamic libraries required to connect to an external database on the MacOS distribution were not being bundle correctly. I want to thank Claude Code for the code on the fix.


sergiogpajares:
Fixes issue #1113

otahirs:
thanks for the effort 👍 
there are slight issues like the ODBC driver is not used by the app (linux pipeline excludes it deliberately)
tip: when using claude, iterate on the solution, ask it to review/improve/rewrite/simplify its own code
closing in favor of #1141 which should address bundling missing dependencies in a more general way

PR #1161, opened 2026-06-01
Improve support for Vacants in Start Lists reports & IOF XML exports

ljezek:
### Context
Contributes to #322 

We need ability to export all vacants for automatic processing of late entries (i.e., dohlas.se).

### Changes
   * **Add Vacants options in Start Lists**: Added 3 options for start-list exports: Only runners, Vacants with time (current functionality), All vacants (includes vacants in classes with free start time).
   * **Add Vacant exports to Starters Start List**: add the same 3 options to runners start list, reuse the same function for computing vacants, but make the Starters list only handle reformatting.
   * **Refine IOF XML 3.0 start list**: Updated start list XML exports for vacants, specifically omit the `<Person>` element ([IOF.xsd:2019](https://github.com/international-orienteering-federation/datastandard-v3/blob/24eb108e4c6b5e2904e5f8f0e49142e45e2c5230/IOF.xsd#L2019)), always add `<StartTime>` and omit `<BibNumber>`.

### Tests
* Tested locally (see the report in the #322 issue)
* Exported start list with all vacants to ORIS - https://oris.ceskyorientak.cz/Startovka?id=9933#218702

### Out of scope
* Updating start list exporters (OResults/OFeed) will be done in a follow up

PR #1163, opened 2026-06-02
Feature/startersimprove - try to minimize the diffs in the starters list function

ljezek:
This is an attempt to address complexity concerns from https://github.com/Quick-Box/quickevent/pull/1161 (not yet ready for review, just attempt to see the diffs and estimate if it's better or not).


fvacek:
any progress ?

PR #1166, opened 2026-06-10
Feature/relaysmode - hide Relays tab when not relevant 

lukaskett:
Show `Relays` tab only for relays and teams event. Hide also actions (Print, Import, Export) in `Runs` tab.
#972

### DIFF
--- a/libqf/libqfgui/src/framework/partswitch.cpp
+++ b/libqf/libqfgui/src/framework/partswitch.cpp
@@ -47,12 +47,18 @@ void PartSwitch::addPartWidget(PartWidget *widget)
 	});
 	bt->setText(widget->title());
 	bt->setPartIndex(buttonCount());
-	addWidget(bt);
+	m_partActions.append(addWidget(bt));
 	{
 		QIcon ico = widget->createIcon();
 		//bt->setIconSize(QSize{128, 128});
 		bt->setIcon(ico);
 	}
+}
+
+void PartSwitch::setPartVisible(int part_index, bool visible)
+{
+	if(part_index >= 0 && part_index < m_partActions.size())
+		m_partActions[part_index]->setVisible(visible);
 }
 
 void PartSwitch::setCurrentPartIndex(int ix, bool is_active)
@@ -82,21 +88,14 @@ void PartSwitch::setCurrentPartIndex(int ix, bool is_active)
 
 int PartSwitch::buttonCount()
 {
-	QList<PartSwitchToolButton*> lst = findChildren<PartSwitchToolButton*>(QString(), Qt::FindDirectChildrenOnly);
-	return lst.count();
+	return m_partActions.size();
 }
 
 PartSwitchToolButton *PartSwitch::buttonAt(int part_index)
 {
-	PartSwitchToolButton *ret = nullptr;
-	const QList<PartSwitchToolButton*> lst = findChildren<PartSwitchToolButton*>(QString(), Qt::FindDirectChildrenOnly);
-	for(auto bt : lst) {
-		if(bt->partIndex() == part_index) {
-			ret = bt;
-			break;
-		}
-	}
-	return ret;
+	if(part_index < 0 || part_index >= m_partActions.size())
+		return nullptr;
+	return qobject_cast<PartSwitchToolButton*>(widgetForAction(m_partActions[part_index]));
 }
 
 
--- a/libqf/libqfgui/src/framework/partswitch.cpp
+++ b/libqf/libqfgui/src/framework/partswitch.cpp
@@ -47,7 +47,7 @@ void PartSwitch::addPartWidget(PartWidget *widget)
 	});
 	bt->setText(widget->title());
 	bt->setPartIndex(buttonCount());
-	m_partActions.append(addWidget(bt));
+	addWidget(bt);
 	{
 		QIcon ico = widget->createIcon();
 		//bt->setIconSize(QSize{128, 128});
@@ -57,8 +57,14 @@ void PartSwitch::addPartWidget(PartWidget *widget)
 
 void PartSwitch::setPartVisible(int part_index, bool visible)
 {
-	if(part_index >= 0 && part_index < m_partActions.size())
-		m_partActions[part_index]->setVisible(visible);
+	for(auto *a : actions()) {
+		if(auto *bt = qobject_cast<PartSwitchToolButton*>(widgetForAction(a))) {
+			if(bt->partIndex() == part_index) {
+				a->setVisible(visible);
+				break;
+			}
+		}
+	}
 }
 
 void PartSwitch::setCurrentPartIndex(int ix, bool is_active)
@@ -88,14 +94,18 @@ void PartSwitch::setCurrentPartIndex(int ix, bool is_active)
 
 int PartSwitch::buttonCount()
 {
-	return m_partActions.size();
+	return findChildren<PartSwitchToolButton*>(QString(), Qt::FindDirectChildrenOnly).count();
 }
 
 PartSwitchToolButton *PartSwitch::buttonAt(int part_index)
 {
-	if(part_index < 0 || part_index >= m_partActions.size())
-		return nullptr;
-	return qobject_cast<PartSwitchToolButton*>(widgetForAction(m_partActions[part_index]));
+	QList<PartSwitchToolButton*> lst = findChildren<PartSwitchToolButton*>(QString(), Qt::FindDirectChildrenOnly);
+	for(auto *bt : lst) {
+		if(bt->partIndex() == part_index) {
+			return bt;
+		}
+	}
+	return nullptr;
 }
 
 
--- a/libqf/libqfgui/src/framework/stackedcentralwidget.cpp
+++ b/libqf/libqfgui/src/framework/stackedcentralwidget.cpp
@@ -67,6 +67,13 @@ bool StackedCentralWidget::setActivePart(int part_index, bool set_active)
 	return ret;
 }
 
+void StackedCentralWidget::setPartVisible(const QString &feature_id, bool visible)
+{
+	int ix = featureToIndex(feature_id);
+	if(ix >= 0)
+		m_partSwitch->setPartVisible(ix, visible);
+}
+
 int StackedCentralWidget::featureToIndex(const QString &feature_id)
 {
 	for (int i = 0; i < m_centralWidget->count(); ++i) {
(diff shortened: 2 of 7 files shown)


lukaskett:
There is a workaround for Qt 6.8 bug: `delegateForIndex` returns a `QPointer` from its internal map without null-checking it. Deleting the delegate before removing the map entry left a null `QPointer`, which `paintEvent` then dereferenced. The fix was two lines: swap the deletion order in `runstablewidget.cpp`, plus a `TableView::itemDelegateForIndex` override as a safety net for any future occurrence.

Relays
image

Individual event
image

libqf/libqfgui/src/tableview.cpp
```diff
@@ -93,6 +93,28 @@ TableView::TableView(QWidget *parent) :
 
 	m_proxyModel = new TableViewProxyModel(this);
 	connect(m_proxyModel, &TableViewProxyModel::modelReset, this, &TableView::refreshActions);
+	// Qt6 bug workaround: QSortFilterProxyModel clears its internal mapping during reset,
```
fvacek: I'm not sure, why this change is needed?

libqf/libqfgui/src/tableview.cpp
```diff
@@ -93,6 +93,28 @@ TableView::TableView(QWidget *parent) :
...
+			sm->clearCurrentIndex();
+			sm->clearSelection();
+		}
+	});
+	// After reset or row removal, QAbstractItemViewPrivate::hover keeps a stale QModelIndex
+	// whose internalPointer points to freed QSortFilterProxyModel mapping memory.
+	// Sending Leave to the viewport forces Qt to clear d->hover via setHoverIndex(invalid).
+	// Our visualRect override validates the internalPointer before any call reaches Qt's
+	// visualRect, so this Leave event is processed safely even with the stale hover.
+	connect(m_proxyModel, &QAbstractItemModel::modelReset, this, [this]() {
```
fvacek: I'm not sure, why this change is needed?

libqf/libqfgui/src/tableview.cpp
```diff
@@ -109,6 +131,37 @@ TableView::~TableView()
 	savePersistentSettings();
 }
 
+QAbstractItemDelegate *TableView::itemDelegateForIndex(const QModelIndex &index) const
```
fvacek: I'm not sure, why this change is needed?

libqf/libqfgui/src/tableview.cpp
```diff
@@ -109,6 +131,37 @@ TableView::~TableView()
...
+	// QPointer<QAbstractItemDelegate> from its rowDelegates/columnDelegates maps without
+	// null-checking it. If a registered delegate was deleted before its map entry was
+	// removed, the QPointer silently becomes null and paintEvent crashes calling paint().
+	// Guard here and fall back to the generic item delegate (always non-null).
+	if (QAbstractItemDelegate *d = Super::itemDelegateForIndex(index))
+		return d;
+	return itemDelegate();
+}
+
+QRect TableView::visualRect(const QModelIndex &index) const
```
fvacek: I'm not sure, why this change is needed?

fvacek (changes requested):
Please read AI slop carefully before creating PR. Commit only related changes.

arnost00:
> There is a workaround for Qt 6.8 bug

Do we know which Qt bug it was and whether it has been fixed in Qt 6.9, 6.10, 6.11?
Qt 6.8 is not last released version and was [released 08.10.2024](https://wiki.qt.io/Qt_6.8_Release)
We mostly use QT newer than 6.8 (for actions 6.10)

lukaskett:
> > There is a workaround for Qt 6.8 bug
> 
> Do we know which Qt bug it was and whether it has been fixed in Qt 6.9, 6.10, 6.11? Qt 6.8 is not last released version and was [released 08.10.2024](https://wiki.qt.io/Qt_6.8_Release) We mostly use QT newer than 6.8 (for actions 6.10)

 @fvacek is right—I should have double-checked the AI's response before posting. I tried my best, but this was way beyond my expertise. The truth is, it took me quite a bit of effort to get the tab to hide, so something probably isn't working quite right there.  I'll check it again.

arnost00:
I don't really understand why there are some changes, and it seems strange to me to fix a Qt bug without specifying a specific bug number, because then the modification cannot be deleted after it has been fixed in Qt.

The comments mention Qt6.8 or just Qt6 - the Qt6 library has many versions with many bugs, fixing some bugs without context doesn't make sense to me. In Qt6.10, all the ones addressed in the PR may be fixed - but we don't know that because we don't know the specific bugs.

For example, list of fixes in some Qt6.8 versions  : 
 - https://code.qt.io/cgit/qt/qtreleasenotes.git/about/qt/6.8.3/release-note.md#fixes
 - https://code.qt.io/cgit/qt/qtreleasenotes.git/about/qt/6.8.2/release-note.md#fixes

arnost00:
> The truth is, it took me quite a bit of effort to get the tab to hide, so something probably isn't working quite right there. I'll check it again.

Please try using Qt6.10 or 6.11 and not an older version. We are currently using Qt6.10, the next step is Qt6.11.

arnost00:
Wouldn't it be easier and faster to just disable those elements instead of hiding them?

For example `m_actPrint->setEnabled(!is_relays);`

libqf/libqfgui/src/tableview.cpp
lukaskett: removed, thank you

libqf/libqfgui/src/tableview.cpp
lukaskett: removed, thank you

libqf/libqfgui/src/tableview.cpp
lukaskett: removed, thank you

libqf/libqfgui/src/tableview.cpp
lukaskett: removed, thank you

quickevent/app/quickevent/plugins/Runs/src/runswidget.cpp
```diff
@@ -72,7 +73,10 @@ RunsWidget::RunsWidget(QWidget *parent) :
 		ui->cbxDrawMethod->clear();
 		if(!getPlugin<EventPlugin>()->isEventOpen())
 			return;
-		bool is_relays = getPlugin<EventPlugin>()->eventConfig()->isRelays();
+		auto *event_cfg = getPlugin<EventPlugin>()->eventConfig();
+		if(!event_cfg)
+			return;
+		bool is_relays = event_cfg->isRelays();
```
arnost00: Why not use new function `isRelaysEvent()` ?

arnost00:
obrazek

Local build on Qt6.11.1

libqf/libqfgui/src/framework/stackedcentralwidget.cpp
```diff
@@ -7,27 +7,31 @@
...
+	m_partSwitch->setMovable(false);
 	connect(this, &StackedCentralWidget::partActivated, m_partSwitch, &PartSwitch::setCurrentPartIndex);
-	MainWindow::frameWork()->addToolBar(Qt::LeftToolBarArea, m_partSwitch);
+
+	QBoxLayout *ly = new QHBoxLayout(this);
+	ly->setContentsMargins(0, 0, 0, 0);
+	ly->setSpacing(0);
+	ly->addWidget(m_partSwitch);
+	ly->addWidget(m_centralWidget, 1);
+	setLayout(ly);
```
arnost00, line 26: It looks like the layout is broken here.

arnost00:
Why not use `act->setDisabled(xyz)` for `PartSwitch` to only disable module ?

libqf/libqfgui/src/framework/stackedcentralwidget.cpp
lukaskett: Should be fixed. Please, test it.

lukaskett:
> Why not use `act->setDisabled(xyz)` for `PartSwitch` to only disable module ?

In my opinion, it doesn't make sense to display a tab that's been disabled and isn't relevant to a specific race at all. I decided on a combination: the tab is hidden and the actions are disabled. I hope you’ll accept 🙂 .

lukaskett:
PR also fixes the bug when switching from individual race to relays, the app crashes at this moment.

arnost00:
Open relays race, switch to module Relays, open single race and :
obrazek

lukaskett:
> Open relays race, switch to module Relays, open single race and : obrazek

I know, corner case in my point of view. Which tab do you suggest as default?

arnost00:
> I know, corner case in my point of view. Which tab do you suggest as default?

I think we can use  for all type of races `Runs` as default module.

lukaskett:
> I think we can use for all type of races `Runs` as default module.

Should be fixed.

fvacek (changes requested):
I'm afraid that AI have introduced too many unrelated changes that review is much more difficult. I'm sure that human can solve this task much more simpler way. Sorry.

libqf/libqfgui/src/framework/partswitch.cpp
```diff
@@ -82,21 +88,14 @@ void PartSwitch::setCurrentPartIndex(int ix, bool is_active)
 
 int PartSwitch::buttonCount()
 {
-	QList<PartSwitchToolButton*> lst = findChildren<PartSwitchToolButton*>(QString(), Qt::FindDirectChildrenOnly);
```
fvacek, line 85: what is wrong with children introspection, that `m_partActions` must be introduced?

quickevent/app/quickevent/plugins/Runs/src/runswidget.cpp
```diff
@@ -422,6 +422,17 @@ void RunsWidget::settleDownInPartWidget(::PartWidget *part_widget)
 		bt->setCheckable(true);
 		connect(bt, &QPushButton::toggled, ui->frmDrawing, &QFrame::setVisible);
 	}
+
+	auto updateMenuEnabled = [part_widget]() {
+		auto *ep = getPlugin<EventPlugin>();
+		auto *cfg = ep->isEventOpen() ? ep->eventConfig() : nullptr;
+		bool enabled = !(cfg && cfg->isRelays());
+		for(const char *path : {"print", "import", "export"})
```
fvacek: always use curly braces, please

fvacek (changes requested):

libqf/libqfgui/src/framework/partswitch.cpp
lukaskett: The `m_partActions` member is not necessary, thanks for pointing out.

quickevent/app/quickevent/plugins/Runs/src/runswidget.cpp
lukaskett: updated

fvacek (approved):

PR #1168, opened 2026-06-11
Runs: add prev/next class navigation buttons

tkafka:
## Summary
Adds **`<`** and **`>`** toolbar buttons next to the class combo box in the Runs/Stages widget, so classes can be stepped through quickly for visual checking.

## Details
- Navigation steps only through real classes — the `--- all ---` pseudo-item is skipped.
- Stepping is clamped via `qBound`, so navigating past the first or last class is a safe no-op (cannot crash).
- The buttons are disabled at the boundaries (on the first / last class).
- Czech translations for the two new tooltips are included.

## Testing
- Full CMake build (Qt 6.11.1 / MinGW 13.1) passes with no errors or warnings.
- Verified by running the app and stepping through classes in the Runs/Stages panel.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/quickevent/app/quickevent/plugins/Runs/src/runswidget.cpp
+++ b/quickevent/app/quickevent/plugins/Runs/src/runswidget.cpp
@@ -36,6 +36,7 @@
 #include <QLabel>
 #include <QInputDialog>
 #include <QTimer>
+#include <QToolButton>
 #include <QRandomGenerator>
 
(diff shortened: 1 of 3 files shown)


fvacek (approved):

lukaskett:
I noticed that it's missing in `Relays` tab.

PR #1169, opened 2026-06-11
Draw tool: visualize manually entered start times, conflicts sidebar, slot header spinners

tkafka:
## Česky:

Co jsme udělali s Nástrojem pro losování (Ctrl+L)

1. Respektuje ručně zadané starty. Dřív položil všechny kategorie na čas 0 a hodnoty z tabulky ignoroval. Teď si můžu zadat start a intervaly v tabulce a pak přepnout do vizualizace, doladi a uložit.

2. Nová hlavička řádku - umí editovat jak start, tak interval, a u obojího ukáže i < a > tlačítka pro přesné posouvání. Cíl: objevitelnost a použitelnost i např. pro touchpad.

3. Sidebar konfliktů. Nově se nepočítá jen první kolize, ale v pravém sidebaru se zobrazí všechny najednou, a můžu si je vybrat a vyřešit. Kdo nechce sidebar používat, může si ho vypnout, gantt chart stejně vždycky ukáže jen první nevyřešený konflikt (kvůli přehlednosti).

4. Ochrana dat. Zavření okna s neuloženými změnami se zeptá Uložit / Zahodit / Zrušit. Po cestě jsme opravili i bug, kdy Edit class nepřepočítal po uložení kolize.

Vytvořeno Claudem, ale všechny funkce jsem otestoval.

## Náhled:

image

## Summary

This PR makes the draw tool (Classes → Edit → Classes layout, Ctrl+L) usable for a workflow where class start times are entered manually in the Classes table first, and the draw tool is then opened to review and fine-tune the layout. Previously the tool ignored manually entered start times and placed all classes not laid out by the tool itself at time 0.

### Load respects the Classes table

- `classdefs.startTimeMin` is now authoritative on load: every start slot is positioned by the start time of its first class, and classes never laid out by the tool get their own slot at their manually entered start time.
- Consecutive same-course classes stay chained in one slot when they continue without a gap (or have no start time set yet), so the original draw-from-scratch workflow is unchanged.
- The Edit class dialog opened from the tool now recomputes clashes and whole-canvas geometry, and moves the slot when the start time of its first class was changed.

### Slot header redesigned as a 2×2 grid

- Column 1: slot number (gray, centered) above the clash check lock (square hover/click area).
- Column 2: `◀ value ▶` spinners for the slot start time and the start interval, with hover highlight, mouse wheel support and ±1 minute buttons.
- The interval spinner sets the interval of all classes in the slot, so `GanttItem::save()` now also persists `startIntervalMin`.
- Empty slots used to be drawn smaller; row height is now defined once in `GanttScene::rowHeight()` and shared by class items and slot headers.

### Conflicts sidebar

- `GanttItem::checkClassClash()` now collects **all** clashing class pairs (it used to stop at the first one) and publishes them via a new `GanttScene::clashesChanged()` signal.
- A toggleable sidebar lists every conflict ("A × B — same course N" / "A × B — first control N"); clicking an item highlights that conflict and scrolls it into view.
- Exactly one conflict is highlighted in the visualization at a time — the first unresolved one by default — so the tool behaves like before when the sidebar is not used, and the selection survives layout edits until the conflict is resolved.
- A "Show all conflicts (N)" checkbox with a live count sits at the right end of the toolbar, above the sidebar.

### Unsaved changes guard

- Closing the dialog with unsaved layout changes asks to save / discard / cancel (via `DialogWidget::acceptDialogDone`). All layout mutations mark the scene dirty; load and save reset it.

### Tooltips & translations

- Fixed the misleading lock tooltip ("Lock class start time" — it actually toggles clash checking) and added tooltips to slot headers, class tiles, toolbar actions and clash check boxes.
- Czech translations added for all new strings; other languages will pick them up via lupdate.

## Test plan

- [x] Open the draw tool on an event with manually entered start times — classes appear at their start times instead of 0
- [x] Adjust slot start/interval via spinners, mouse wheel and buttons; clashes recompute live
- [x] Edit class via right-click; clashes and geometry recompute, slot follows the new start time
- [x] Conflicts sidebar lists all conflicts; clicking highlights and scrolls; resolving one selects the next
- [x] Closing with unsaved changes prompts Save/Discard/Cancel; clean open+close does not prompt
- [x] Save → reopen round-trips the layout unchanged
- [x] Built with Qt 6.11.1 / MinGW 13.1 on Windows, no new warnings

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/quickevent/app/quickevent/plugins/Classes/src/drawing/drawingganttwidget.cpp
+++ b/quickevent/app/quickevent/plugins/Classes/src/drawing/drawingganttwidget.cpp
@@ -128,7 +128,7 @@ bool DrawingGanttWidget::acceptDialogDone(int result)
 		auto answer = QMessageBox::question(this, tr("Unsaved changes"),
 											tr("The start times layout has unsaved changes.\n"
 											   "Do you want to save them before closing?"),
-											QMessageBox::Save | QMessageBox::Discard | QMessageBox::Cancel,
+											QMessageBox::Discard | QMessageBox::Cancel,
 											QMessageBox::Save);
 		if(answer == QMessageBox::Cancel)
 			return false;
--- a/quickevent/app/quickevent/plugins/Classes/src/drawing/drawingganttwidget.cpp
+++ b/quickevent/app/quickevent/plugins/Classes/src/drawing/drawingganttwidget.cpp
@@ -12,6 +12,7 @@
 #include <qf/gui/dialogs/dialog.h>
 
 #include <QLineEdit>
+#include <QListWidget>
 #include <QMessageBox>
 #include <QCheckBox>
 
@@ -36,6 +37,16 @@ DrawingGanttWidget::DrawingGanttWidget(QWidget *parent) :
 
 	m_ganttScene = new GanttScene(this);
 	ui->ganttView->setScene(m_ganttScene);
+
+	connect(m_ganttScene, &GanttScene::clashesChanged, this, &DrawingGanttWidget::onClashesChanged);
+	connect(ui->lstClashes, &QListWidget::itemClicked, this, &DrawingGanttWidget::onClashItemActivated);
+	connect(ui->lstClashes, &QListWidget::currentItemChanged, this, [this](QListWidgetItem *current, QListWidgetItem *) {
+		if(current)
+			onClashItemActivated(current);
+	});
+	ui->lstClashes->setToolTip(tr("Click a conflict to highlight it in the layout"));
+	ui->splitter->setStretchFactor(0, 1);
+	ui->splitter->setStretchFactor(1, 0);
 }
 
 DrawingGanttWidget::~DrawingGanttWidget()
@@ -89,12 +100,40 @@ void DrawingGanttWidget::settleDownInDialog(qf::gui::dialogs::Dialog *dlg)
 		connect(cb, &QCheckBox::checkStateChanged, this, update_class_check);
 #endif
 	}
+	{
+		// push the conflict list checkbox to the right, above the conflicts sidebar
+		auto *spacer = new QWidget();
+		spacer->setSizePolicy(QSizePolicy::Expanding, QSizePolicy::Preferred);
+		tb->addWidget(spacer);
+		m_cbShowClashes = new QCheckBox(tr("Show all conflicts"));
+		m_cbShowClashes->setChecked(true);
+		m_cbShowClashes->setToolTip(tr("Show the list of start time conflicts"));
+		connect(m_cbShowClashes, &QCheckBox::toggled, ui->lstClashes, &QWidget::setVisible);
+		tb->addWidget(m_cbShowClashes);
+	}
 	update_class_check();
 
 	//auto *menu = dlg->menuBar();
 	//auto *a_draw = menu->actionForPath("draw");
 	//a_draw->setText(tr("&Draw"));
 	//a_draw->addActionInto(ui->actSave);
+}
+
+bool DrawingGanttWidget::acceptDialogDone(int result)
+{
+	Q_UNUSED(result)
+	if(m_ganttScene->isDirty()) {
+		auto answer = QMessageBox::question(this, tr("Unsaved changes"),
+											tr("The start times layout has unsaved changes.\n"
+											   "Do you want to save them before closing?"),
+											QMessageBox::Save | QMessageBox::Discard | QMessageBox::Cancel,
+											QMessageBox::Save);
+		if(answer == QMessageBox::Cancel)
+			return false;
+		if(answer == QMessageBox::Save)
+			m_ganttScene->save();
+	}
+	return true;
 }
 
 void DrawingGanttWidget::load(int stage_id)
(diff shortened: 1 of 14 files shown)


tkafka:
Two follow-up commits based on further testing:

**Start interval spinner is disabled for slots with mixed intervals** (`df535113`). Previously the spinner showed the first class's interval and silently unified all classes in the slot on the first edit. Now, when classes in a slot have different start intervals, the spinner is grayed out, shows `≠`, and its tooltip lists the classes with their intervals (e.g. `D10F: 4, H10F: 4, HDR: 0`) so the mismatch is easy to track down. It re-enables as soon as the slot is uniform again.

**Zero-duration classes are now visible** (`df535113`). Classes with no start interval have zero width and used to be completely invisible in the gantt while still occupying a slot — typically an HDR-style free-start class chained behind its course mates, which made the slot look mysteriously "blocked". They are now painted as a narrow solid flag: `boundingRect()`/`shape()` are widened to a few pixels so the class can be hovered (tooltip identifies it), dragged to another slot and right-clicked, while its layout width stays zero so start times of the following classes are unaffected.

Also `ad4b80fc` fixes the clang-tidy findings from CI (static `Style::icon()` access, `ClassItem::clashWith()` constness, dead store in the slot header drag).

One clarification to the description: *"classdefs.startTimeMin is now authoritative on load"* — start **intervals** were always read from `classdefs.startIntervalMin` on load (they have no competing copy in `drawingConfig`); what is new for them is the opposite direction, i.e. Save now writes interval edits made in the tool back to the Classes table.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

quickevent/app/quickevent/plugins/Classes/src/drawing/drawingganttwidget.cpp
```diff
@@ -82,6 +100,17 @@ void DrawingGanttWidget::settleDownInDialog(qf::gui::dialogs::Dialog *dlg)
 		connect(cb, &QCheckBox::checkStateChanged, this, update_class_check);
 #endif
 	}
+	{
+		// push the conflict list checkbox to the right, above the conflicts sidebar
+		auto *spacer = new QWidget();
```
fvacek, line 107: Why we need empty widget?

quickevent/app/quickevent/plugins/Classes/src/drawing/ganttscene.h
```diff
@@ -20,23 +24,33 @@ class GanttScene : public QGraphicsScene
 
 	GanttItem *ganttItem() { return m_ganttItem; }
 
+	void notifyClashesChanged(const QList<ClassClash> &clashes) { emit clashesChanged(clashes); }
```
fvacek:
```suggestion
	void emitClashesChanged(const QList<ClassClash> &clashes) { emit clashesChanged(clashes); }
```

fvacek (changes requested):
I'm not sure, if all the classes classes can be found. Previous implementation had found only first clash. Can you explain, how do you find all the classes clashes?

quickevent/app/quickevent/plugins/Classes/src/drawing/drawingganttwidget.cpp
tkafka: The empty widget is just a spacer. QToolBar has no addStretch() (unlike QBoxLayout), so the idiomatic way to push the following item to the right edge of the toolbar is an empty QWidget with an Expanding horizontal size policy. Kept it and added a comment explaining this in 9af630dd.

quickevent/app/quickevent/plugins/Classes/src/drawing/ganttscene.h
tkafka: Done in 9af630dd — renamed to emitClashesChanged() (and its one caller in ganttitem.cpp).

lukaskett:
@tkafka could you please extend the PR with #747? It's "only" about the initial sort. Thank you for considering this 🙂 .

quickevent/app/quickevent/plugins/Classes/src/drawing/drawingganttwidget.cpp
```diff
@@ -90,6 +121,23 @@ void DrawingGanttWidget::settleDownInDialog(qf::gui::dialogs::Dialog *dlg)
...
 }
 
+bool DrawingGanttWidget::acceptDialogDone(int result)
+{
+	Q_UNUSED(result)
+	if(m_ganttScene->isDirty()) {
+		auto answer = QMessageBox::question(this, tr("Unsaved changes"),
+											tr("The start times layout has unsaved changes.\n"
+											   "Do you want to save them before closing?"),
+											QMessageBox::Save | QMessageBox::Discard | QMessageBox::Cancel,
```
fvacek:
```suggestion
											QMessageBox::Discard | QMessageBox::Cancel,
```

fvacek (commented):
Sorry for late answer, I've tested proposed changes and it looks like very useful drawing tool enhancement.

PR #1173, opened 2026-06-16
fix: missing Date element value in both startlist and results for relays

lukaskett:
The same issue as for individual race that was fixed in #1128.

Fixes issue #1127.


arnost00:
It looks good.

arnost00 (approved):

fvacek (approved):

PR #1180, opened 2026-07-16
801 event edit can also be moved to settings   event

### LINKED ISSUE #801: 1. Event edit can also be moved to settings -> `Event`
1. Event edit can also be moved to settings -> `Event`
2. I also think that `Check type - Classic race | Free order race` should be moved into this form
image

_Originally posted by @otahirs in https://github.com/Quick-Event/quickbox/issues/780#issuecomment-1231388611_

fvacek:
#801

- Moved event configuration to structured DB-backed `AppDbConfig`.
- Replaced dynamic/QObject `EventConfig` usage with typed config structs.
- Added typed configs for stages, OFeed, OResults, Receipts, and QX.
- Changed config setters to persist only changed fields.
- Moved stage start times and stage settings into event config.
- Removed old stage editor widget/document/data classes.
- Restored creation of `stages` table rows for FK compatibility.
- Reworked current-stage UI with toolbar/status-bar integration.
- Added “Set current stage” action and stage chooser flow.
- Renamed `EventPlugin` event identifier usage to `eventDbName`.
- Updated services to use the new event/stage config APIs.
- Made OFeed and OResults configuration per-stage.
- Moved receipt QR/image settings into typed per-stage config.
- Fixed receipt QR link URL serialization.
- Updated Runs, Classes, CardReader, Relays, Oris, XML import/export paths for typed config access.
- Updated report award templates for map-style `eventConfig` access.
- Updated build files to include new config sources and remove old stage editor files.
 
image


fvacek:
@sergiogpajares

lukaskett:
Good job, it simplifies the UI a lot. Just two small issues i noticed while testing:
- We could remove the top element with stage selection (i tried to move it) and save space as you mentioned earlier. The same selector is available in the lower roght corner and as an option in `Event` menu.
image

- I noticed wrong 1st stage number - shows date instead of simple number. Looks like an issue with data formating or column width and it's `Windows` specific.
  - COT (2 stages)
image

  - Rumile (3 stages)
image

fvacek:
> We could remove the top element with stage selection (i tried to move it) and save space as you mentioned earlier. The same selector is available in the lower roght corner and as an option in Event menu.

I was thinking about this, but I still feel, that current stage is important information and should be highly visible. I prefer this location 
image

sergiogpajares:
Sorry for the very late response. I've been a bit overwhelmed with some personal matters and orienteering commitments. I also won't be very active during August.

I think this is a very nice improvement. Thank you very much for taking the proposal into consideration.

PR #1188, opened 2026-08-05
Fix IOF-XML results start time only if not set

### LINKED ISSUE #1187: Bug in IOF 3.0 XML result export if is used finish gate
QE 3.6.0 - 3.6.3
If is used finish gate, export result to XML format export incorect start time.
Because inner format is msec and export only sec.

XML:
```
<StartTime>2026-08-05T12:44:59</StartTime>
<FinishTime>2026-08-05T13:38:08</FinishTime>
```
QE:
```
Start   	165.00	-> 12:45:00
Finish SI  	218.08
Finish Gate	218.08/697
```
In XML is time 1 second lower.

The problem is in code `runsplugin.cpp` line 1113 and 1114
```
			if(ftime && time)
				stime = ftime - time; // cover cases when competitor didn't started according to start list from any reason

``` 
Which calculate time lower then 165.00 (164.59/303) and it is then trimmed to 164.59 (12:44:59)

---

Result time is not affected.

fvacek:


arnost00:
Testing XML result export looks good.

arnost00 (approved):

PR #1189, opened 2026-08-06
1101 generate point based results

### LINKED ISSUE #1101: Generate point-based results
In Spain, it is common to use the 1000-points formula (i.e.  1000 * best time / runner’s time)

fvacek:
#1101 

image

Image

Image

### DIFF
--- a/libqf/libqfcore/src/utils/table.cpp
+++ b/libqf/libqfcore/src/utils/table.cpp
@@ -1150,9 +1150,9 @@ QString Table::toString() const
 	QString s;
 	QTextStream ts(&s);
 	//ts.setPadChar('%');
-	ts << "       ";
+	ts << "   #   ";
 	for(int i=0; i<fields().count(); i++) {
-		if(i > 0) ts << sep;
+		ts << sep;
 		ts << field(i).name();
 	}
 	ts << '\n';
@@ -1174,12 +1174,12 @@ QString Table::toString() const
 
 qf::core::Collator Table::sortCollator() const
 {
-    return d->sortCollator;
+	return d->sortCollator;
 }
 
 void Table::setSortCollator(const Collator &coll)
 {
-    d->sortCollator = coll;
+	d->sortCollator = coll;
 }
 
 void Table::sort(const QString &_colnames)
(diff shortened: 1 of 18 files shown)


arnost00 (approved):
Look good.
But maybe @sergiogpajares can check it better.

Only small UI note, tab Import/Export in Edit Event uses only like 1/4 of tab area 

Image

lukaskett:
What about skipping one stage/two stages/the worst result? I know that there can be e.g. 5 stages event but only 4 best stages/results are counted into the overall results.

fvacek:
> Look good. But maybe @sergiogpajares can check it better.
> 
> Only small UI note, tab Import/Export in Edit Event uses only like 1/4 of tab area
> Image

Yes, this is IMO better than have number spin box as wide as the dialog.

fvacek:
> What about skipping one stage/two stages/the worst result? I know that there can be e.g. 5 stages event but only 4 best stages/results are counted into the overall results.

Yes, this requirement is part of issue already https://github.com/Quick-Box/quickevent/issues/1101

lukaskett:
> > What about skipping one stage/two stages/the worst result? I know that there can be e.g. 5 stages event but only 4 best stages/results are counted into the overall results.
> 
> Yes, this requirement is part of issue already #1101

Where can i set which stage should be calculated and which not?
image

I was not able to find that option. The settings has just `Max points` configuration.
image

and stages have just start time attribute stored
image

fvacek:
> > > What about skipping one stage/two stages/the worst result? I know that there can be e.g. 5 stages event but only 4 best stages/results are counted into the overall results.
> > 
> > 
> > Yes, this requirement is part of issue already #1101
> 
> Where can i set which stage should be calculated and which not? image
> 
> I was not able to find that option. The settings has just `Max points` configuration. image
> 
> and stages have just start time attribute stored image

this feature is not here yet https://github.com/Quick-Box/quickevent/issues/1101#issuecomment-5245615258

lukaskett:
> > > > What about skipping one stage/two stages/the worst result? I know that there can be e.g. 5 stages event but only 4 best stages/results are counted into the overall results.
> > > 
> > > 
> > > Yes, this requirement is part of issue already #1101
> > 
> > 
> > Where can i set which stage should be calculated and which not? image
> > I was not able to find that option. The settings has just `Max points` configuration. image
> > and stages have just start time attribute stored image
> 
> this feature is not here yet [#1101 (comment)](https://github.com/Quick-Box/quickevent/issues/1101#issuecomment-5245615258)

ou, i finally got it 😄 PR title confused me...

PR #1198, opened 2026-08-13
Searching relay runners by Bib, when using assign card to runner.

VojtechVelat:
When organizing a relay it is not possible to search runners by Bib, when using the "assign card to runner". All bibs default to 0.
image

This commit fixes this problem.
image

### DIFF
--- a/aqtinstall.log
+++ b/aqtinstall.log
@@ -1,0 +1,242 @@
+2026-08-12 23:49:04,433 - aqt.main - INFO - installer 25164 aqtinstall(aqt) v3.3.0 on Python 3.12.6 [CPython MSC v.1940 64 bit (AMD64)]
+2026-08-12 23:49:04,433 - aqt.helper - DEBUG - helper 25164 Attempt to download checksum at https://download.qt.io/online/qtsdkrepository/windows_x86/desktop/tools_mingw1310/Updates.xml.sha256
+2026-08-12 23:49:05,798 - aqt.helper - DEBUG - helper 25164 Attempt to download checksum at https://download.qt.io/online/qtsdkrepository/windows_x86/desktop/tools_mingw1310/Updates.xml.sha256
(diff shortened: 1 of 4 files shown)


aqtinstall.log
fvacek, line 1: should not be part of commit IMO

quickevent/app/quickevent/plugins/Runs/src/runsplugin.cpp
```diff
@@ -79,16 +79,25 @@ RunsPlugin::~RunsPlugin() = default;
...
-				.select2("runs", "id, siId, competitorId")
+				.select2("runs", "id, siId, competitorId, leg")
 				.select("runs.id AS runId")
 				.select2("classes", "name")
 				.from("competitors")
-				.join("competitors.classId", "classes.id")
 				.joinRestricted("competitors.id", "runs.competitorId", "runs.stageId=" QF_IARG(stage_id), "JOIN")
 				.orderBy("classes.name, lastName, firstName");
+		if(is_relays) {
+			qb.select("relays.number AS \"competitors.startNumber\"")
```
fvacek, line 93:
```suggestion
			qb.select("relays.number AS startNumber")
```

quickevent/app/quickevent/plugins/Runs/src/runsplugin.cpp
```diff
@@ -79,16 +79,25 @@ RunsPlugin::~RunsPlugin() = default;
...
+					.join("runs.relayId", "relays.id")
+					.join("relays.classId", "classes.id");
+		}
+		else {
+			qb.select2("competitors", "startNumber")
```
fvacek, line 98:
```suggestion
			qb.select("competitors.startNumber AS startNumber")
```

fvacek (changes requested):

fvacek (commented):
any progress?

